PR #7, opened 2013-12-21
Pull request 5

vitsalis:


controllers/image.php
```diff
@@ -16,7 +16,7 @@ public static function create( $image ) {
                 $user->save();
             }
             catch ( ModelValidationException $e ) {
-                go( 'user', 'update', array( 'username' => $user->username, $e->error => true ) );
+                go( 'user', 'update', array( $e->error => true ) );
```
dionyziz, line 19: I don't understand this change.

database/db_schema.sql
```diff
@@ -28,10 +28,10 @@ SET time_zone = "+00:00";
 
 CREATE TABLE IF NOT EXISTS `countries` (
   `id` int(4) NOT NULL AUTO_INCREMENT,
-  `name` text COLLATE utf8mb4_unicode_ci NOT NULL,
-  `shortname` text COLLATE utf8mb4_unicode_ci NOT NULL,
+  `name` text COLLATE utf8_unicode_ci NOT NULL,
+  `shortname` text COLLATE utf8_unicode_ci NOT NULL,
   PRIMARY KEY (`id`)
-) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci ;
+) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci ;
```
dionyziz, line 34: I've already made these changes, please pull.

controllers/image.php
vitsalis, line 19: You can update a picture in the edit settings page. The edit settings page doesn't need a username as a resource.
When upload image was on the profile it was needed because we needed to know whose profile was that we were visiting.

PR #25, opened 2014-02-01
Country testing

pkakelas:


tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+        protected static function insertCountry() {
+            $country = array(
+               'name' => 'Greece', 
+               'shortname' => 'Gr'
+            );
+            dbInsert( 'countries', $country );
+        }
+        public function testCountryConstruct() {
+            $insert = $created = true;
+            $randomId = 1;
```
vitsalis: why do you call it randomId if it's not random?

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            try {
+                CountryTest::insertCountry();
+            }
+            catch ( DBException $e ) {
+               $insert = false; 
+            }
```
vitsalis, line 18: You are not testing dbInsert, this doesn't belong here

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            try {
+                $country = new Country( $randomId );
+            }
+            catch ( ModelNotFoundException $e ) {
+                $created = false;
+            }
```
vitsalis, line 27: You are should not test if a country is on the database here.
You could make another test for this, like testFindNonExistentCountry()

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            $id = isset( $country->id );
```
vitsalis: why call it $id if it is going to be only true or false?

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            $shortname = isset( $country->shortname );
```
vitsalis: why call it $shortname if it is going to be only true or false?

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            $name = isset( $country->shortname );
```
vitsalis: why call it $name if it is going to be only true or false?
$country->shortname? doesn't $shortname already hold that value?

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            $this->assertEquals( true, $insert, 'Problem with dbInsert. We cannot insert countries in the db' );
```
vitsalis: You should not test dbInsert here

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            $this->assertEquals( 
+                true, 
+                $created, 
+                'Problem with the dbSelectOne() function or there is not any import with this id in the db' 
+            );
```
vitsalis: You should not test dbSelectOne here.
"or"? unit testing should identify the error not give you possibilities.

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            $this->assertEquals( true, $id, 'country->id of the new country object is empty' );
```
vitsalis: You should test if the $id is equal to the $randomId.

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            $this->assertEquals( true, $shortname, 'country->shortname of the new country object is empty' );
```
vitsalis: You should test if the $shortname is equal to the $shortname you gave on insertCountry().

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            $this->assertEquals( true, $id, 'country->name of the new country object is empty' );
```
vitsalis: $id?
If you ment $name, then you should test if the $name is equal to the $name you gave on insertCountry().

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+        }
+        public function testFindAll() {
+            $insert = true;
+            $randomId = 2;
+            try {
+                CountryTest::insertCountry();
+            }
+            catch ( DBException $e ) {
+               $insert = false; 
+            }
```
vitsalis: You're not testing dbInsert(). You should suppose that dbInsert() works just fine.

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+                $created, 
+                'Problem with the dbSelectOne() function or there is not any import with this id in the db' 
+            );
+            $this->assertEquals( true, $id, 'country->id of the new country object is empty' );
+            $this->assertEquals( true, $shortname, 'country->shortname of the new country object is empty' );
+            $this->assertEquals( true, $id, 'country->name of the new country object is empty' );
+        }
+        public function testFindAll() {
+            $insert = true;
+            $randomId = 2;
```
vitsalis: it is not random, so why call it $randomId?

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            try {
+                CountryTest::insertCountry();
+            }
+            catch ( DBException $e ) {
+               $insert = false; 
+            }
+            $country = new Country( $randomId ); 
+            $array = Country::findAll(); 
+            $success = isset( $array );
```
vitsalis: I think it would be better to have an

```
 if ( count( $array ) > 0 ) {
      $success = true;
 }
```

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            try {
+                CountryTest::insertCountry();
+            }
+            catch ( DBException $e ) {
+               $insert = false; 
+            }
+            $country = new Country( $randomId ); 
+            $array = Country::findAll(); 
+            $success = isset( $array ); 
+            $countryId = "$country->id";
```
vitsalis: why make `$countryId` a string and not an int?

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            $this->assertEquals( true, $insert, 'Problem with dbInsert. We cannot insert countries in the db' );
```
vitsalis: You should not test dbInsert() here.

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            $this->assertEquals( true, $success, 'dbSelectOne did not work' );
```
vitsalis: You're not testing dbSelectOne(). You're testing if the findAll() function works.

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+            $this->assertEquals( $array[ $randomId - 1 ][ 'id' ], $countryId, 'Not all countries are imported in the database' );
```
vitsalis: you should use intval() instead of making $countryId a string.
http://www.php.net/intval
You're not testing if countries are imported to the database, but if findAll() finds them.

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+               'name' => 'Greece', 
+               'shortname' => 'Gr'
+            );
+            dbInsert( 'countries', $country );
+        }
+        public function testCountryConstruct() {
+            $insert = $created = true;
+            $randomId = 1;
+            try {
+                CountryTest::insertCountry();
```
vitsalis: $this->insertCountry()

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
+<?php
+    include_once 'models/country.php';
+    
+    class CountryTest extends UnitTestWithUser {
+        protected static function insertCountry() {
```
vitsalis: it should not be static.

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
+<?php
+    include_once 'models/country.php';
+    
+    class CountryTest extends UnitTestWithUser {
```
vitsalis: why extends UnitTestWithUser? You are not using buildUser() anywhere.

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+        protected static function insertCountry() {
+            $country = array(
+               'name' => 'Greece', 
+               'shortname' => 'Gr'
+            );
+            dbInsert( 'countries', $country );
```
dionyziz: Don't insert directly, use active record.

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+        }
+        public function testCountryConstruct() {
+            $insert = $created = true;
```
dionyziz: `$inserted`

tests/models/country.php
dionyziz: assertTrue

tests/models/country.php
dionyziz: assertTrue

tests/models/country.php
```diff
@@ -0,0 +1,60 @@
...
+                true, 
+                $created, 
+                'Problem with the dbSelectOne() function or there is not any import with this id in the db' 
+            );
+            $this->assertEquals( true, $id, 'country->id of the new country object is empty' );
+            $this->assertEquals( true, $shortname, 'country->shortname of the new country object is empty' );
+            $this->assertEquals( true, $id, 'country->name of the new country object is empty' );
+        }
+        public function testFindAll() {
+            $insert = true;
```
dionyziz: `$inserted`

tests/models/country.php
dionyziz: You can also use `( int )$countryId`

tests/models/country.php
dionyziz: Get the id as a return value, don't try to predict it.

tests/models/country.php
pkakelas, line 27: We are inserting a country before we construct a country. So if "new Country" fails, the problem will not occur from the non existence of countries in the db.

vitsalis, line 27: that's why we have a catch ( Exception $e ) on here:
https://github.com/VitSalis/endofcodes/blob/master/tests/base.php#L29
For unexpected mistakes. You should expect that dbSelectOne works fine.

tests/models/country.php
pkakelas: why?

vitsalis: Ι think it's a better API

tests/models/country.php
pkakelas: Now that I migrated to active record, I am not able to do that because when I save the user I cannot have his id

tests/models/country.php
```diff
@@ -0,0 +1,56 @@
...
+            $country->shortname = 'GR';
+            $country->name = 'Greece';
+            $country->save();
+            $name = $country->name;
+        }
+        public function testCountryConstruct() {
+            $inserted = $created = true;
+            $testId = 1;
+            try {
+                $this->insertCountry();
```
dionyziz: Test insertion separately

tests/models/country.php
```diff
@@ -0,0 +1,56 @@
...
+                $country = new Country( $testId );
+            }
+            catch ( ModelNotFoundException $e ) {
+                $created = false;
+            }
+            $idExists = isset( $country->id );
+            $shortnameExists = isset( $country->shortname );
+            $nameExists = isset( $country->name );
+            $this->assertTrue( $inserted, 'Problem with country save. We cannot insert countries in the db' );
+            $this->assertTrue( $created, 'Problem with the dbSelectOne() function or there is not any import with this id in the db' );
```
dionyziz: Assume `dbSelectOne()` works. If you want you can write separate unit tests for db.

tests/models/country.php
```diff
@@ -0,0 +1,56 @@
...
+            try {
+                $country = new Country( $testId );
+            }
+            catch ( ModelNotFoundException $e ) {
+                $created = false;
+            }
+            $idExists = isset( $country->id );
+            $shortnameExists = isset( $country->shortname );
+            $nameExists = isset( $country->name );
+            $this->assertTrue( $inserted, 'Problem with country save. We cannot insert countries in the db' );
```
dionyziz: Assume insert works in this test. Test insertion separately.

tests/models/country.php
```diff
@@ -0,0 +1,56 @@
...
+                $this->insertCountry();
+            }
+            catch ( DBException $e ) {
+               $inserted = false; 
+            }
+            $country = new Country( $testId ); 
+            $array = Country::findAll(); 
+            $success = isset( $array ); 
+            $countryId = intval( $array[ $testId - 1 ][ 'id' ] ); 
+            $this->assertTrue( $inserted, 'Problem with country save. We cannot insert countries in the db' );
```
dionyziz: Test saving separately.

tests/models/country.php
```diff
@@ -0,0 +1,56 @@
...
+            $this->assertTrue( $success, 'dbSelectOne did not work' );
```
dionyziz: Test db separately.

tests/models/country.php
```diff
@@ -0,0 +1,56 @@
...
+            $this->assertEquals( $countryId, $country->id, 'Not all countries are imported in the database' );
```
dionyziz: This error description indicates that you're testing population. This is a unit test; test findAll, not populate. Change your error message and tests.

tests/models/country.php
```diff
@@ -0,0 +1,56 @@
...
+        }
+        public function testCountryConstruct() {
+            $inserted = $created = true;
+            $testId = 1;
+            try {
+                $this->insertCountry();
+            }
+            catch ( DBException $e ) {
+               $inserted = false; 
+            }
```
vitsalis, line 18: dbInsert should not be tested on tests/models/country.php
You can make another test for this. That's why we have Unexpected errors, for things like this.

tests/models/country.php
```diff
@@ -0,0 +1,56 @@
...
+            try {
+                $country = new Country( $testId );
+            }
+            catch ( ModelNotFoundException $e ) {
+                $created = false;
+            }
```
vitsalis, line 27: You should suppose that dbInsert() works fine.

tests/models/country.php
pkakelas, line 27: Ι created a new test to check about Country save

tests/models/country.php
pkakelas: Now that I use active record to save the new country, the only variable about the shortname is $country->shortname.

tests/models/country.php
pkakelas: Check the conversation. Dio suggested a new way to check that.

tests/models/country.php
```diff
@@ -0,0 +1,55 @@
...
+            $this->assertEquals ( $testId, $country->id, "'new country()' did not return the appropriate country" );
+            $this->assertTrue( $shortnameExists, 'country->shortname of the new country object is empty' );
+            $this->assertTrue( $nameExists, 'country->name of the new country object is empty' );
+        }
+        public function testCountryCreate() {
+            $created = true;
+            try {
+                $this->insertCountry();
+            }
+            catch ( DBException $e ) {
```
vitsalis: do not catch DBException here. you are not testing dbInsert() in this test. An error from models/db.php should be an unanticipated error. That's why we have this here:
https://github.com/VitSalis/endofcodes/blob/master/tests/base.php#L29

tests/models/country.php
```diff
@@ -0,0 +1,55 @@
...
+        } 
+        public function testCountryConstruct() {
+            $created = true;
+            $testId = 1;
+            $this->insertCountry();
+            $country = new Country( $testId );
+            $idExists = isset( $country->id );
+            $shortnameExists = isset( $country->shortname );
+            $nameExists = isset( $country->name );
+            $this->assertTrue( $idExists, 'country->id of the new country object is empty' );
```
vitsalis: In the next line you are checking if $testId equals $country->id, so why check here if country->id exists?

tests/models/country.php
```diff
@@ -0,0 +1,55 @@
...
+            $this->assertEquals ( $testId, $country->id, "'new country()' did not return the appropriate country" );
+            $this->assertTrue( $shortnameExists, 'country->shortname of the new country object is empty' );
```
vitsalis: do not test if the shortname exists, test if the shortname you gave on insertCountry() equals the value that $country->shortname holds.

tests/models/country.php
```diff
@@ -0,0 +1,55 @@
...
+            $this->assertTrue( $nameExists, 'country->name of the new country object is empty' );
```
vitsalis: do not test if the name exists, test if the name you gave on insertCountry() equals the value that $country->name holds.

tests/models/country.php
```diff
@@ -0,0 +1,55 @@
...
+    class CountryTest extends UnitTest {
+        protected function insertCountry() {
+            $countries = array(
+                'GR' => 'Greece',
+                'UK' => 'England',
+                'RU' => 'Russia'
+            );
+            $shortnames = array_keys( $countries );
+            $names = array_values( $countries );
+            for ( $i = 0; $i < 2; $i++ ) {
```
vitsalis: This should become:

``` php
foreach ( $countries as $shortname => $name ) {
    ....
}
```

tests/models/country.php
```diff
@@ -0,0 +1,55 @@
+<?php
+    include_once 'models/country.php';
+    
+    class CountryTest extends UnitTest {
+        protected function insertCountry() {
```
vitsalis: insertCountries()

tests/models/country.php
```diff
@@ -0,0 +1,55 @@
...
+                $country = new Country();
+                $country->shortname = $shortnames[ $i ];
+                $country->name = $names[ $i ];
+                $country->save();
+            }
+        }
+        public function testFindAll() {
+            $inserted = true;
+            $this->insertCountry();
+            $res = Country::findAll();
```
vitsalis: I think $countriesArray would be a better name.

tests/models/country.php
```diff
@@ -0,0 +1,55 @@
...
+                $country->name = $names[ $i ];
+                $country->save();
+            }
+        }
+        public function testFindAll() {
+            $inserted = true;
+            $this->insertCountry();
+            $res = Country::findAll(); 
+            $arrayExists = is_array( $res );
+            $success = count( $res ) > 3;
```
vitsalis: at insertCountry() you insert 3 countries.
why should count( $res ) be > 3 in order for findAll() to be succesful?

tests/models/country.php
```diff
@@ -0,0 +1,55 @@
...
+            $this->assertTrue( $arrayExists, '$Country::findall() did not return an array' );
+            $this->assertTrue( $success, '$Country::findall() did not return the as much countries as it should' );
```
vitsalis: "the as much" ?

tests/models/country.php
```diff
@@ -0,0 +1,55 @@
...
+        } 
+        public function testCountryConstruct() {
+            $created = true;
```
vitsalis: you do not use this variable anywhere in this function.

tests/models/country.php
pkakelas: I don't get you

tests/models/country.php
pkakelas: Because, the country table after the template population script will have at least 3 records. I could use = 3 but we'll have to truncate the table after each test and we don't want that.

tests/models/country.php
```diff
@@ -0,0 +1,46 @@
...
+            $countries = $this->countries;
+            $shortnames = array_keys( $countries );
+            $names = array_values( $countries );
+            foreach ( $countries as $shortname => $name ) {
+                $country = new Country();
+                $country->shortname = $shortname;
+                $country->name = $name;
+                $country->save();
+            }
+            return false;
```
dionyziz: Why are you returning false?

tests/models/country.php
```diff
@@ -0,0 +1,46 @@
...
+        }
+        public function testFindAll() {
+            $inserted = true;
```
dionyziz: This is unused as @VitSalis said in the previous review.

tests/models/country.php
```diff
@@ -0,0 +1,46 @@
...
+            $this->insertCountries();
+            $countriesArray = Country::findAll(); 
+            $arrayExists = is_array( $countriesArray );
+            $success = count( $countriesArray ) > 3;
```
dionyziz: Why > 3? See @VitSalis previous comment in his review.

tests/models/country.php
```diff
@@ -0,0 +1,46 @@
...
+            $this->assertTrue( $arrayExists, '$Country::findall() did not return an array' );
```
dionyziz: better written as is_array( $countriesArray ) here directly

tests/models/country.php
```diff
@@ -0,0 +1,46 @@
...
+            $this->assertTrue( $success, '$Country::findall() did not return the as much countries as it should' );
```
dionyziz: better written as an inequality here directly

tests/models/country.php
```diff
@@ -0,0 +1,42 @@
...
+        public function testCountryConstruct() {
+            $countries = $this->countries;
+            $testId = 1;
+            $shortnames = array_keys( $countries );
+            $shortname = $shortnames[ 0 ];
+            $this->insertCountries();
+            $country = new Country( $testId );
+            $this->assertEquals ( $testId, $country->id, "'new country()' did not return the appropriate country" );
+            $this->assertEquals ( $country->shortname, $shortname, "'new country()' did not return the appropriate shortname" );
+            $this->assertEquals ( $country->name, $countries[ $shortname ], "'new country()' did not return the appropriate name" );
```
dionyziz: Coding style in all 3 lines above (should not have space before `(`)

tests/models/country.php
```diff
@@ -0,0 +1,42 @@
...
+                $country->shortname = $shortname;
+                $country->name = $name;
+                $country->save();
+            }
+        }
+        public function testFindAll() {
+            $this->insertCountries();
+            $countriesArray = Country::findAll(); 
+            $this->assertTrue( is_array( $countriesArray ), '$Country::findall() did not return an array' );
+            $this->assertTrue( count( $countriesArray ) > 3, '$Country::findall() did not return the as much countries as it should' );
```
dionyziz: `>=`

dionyziz: `$Country` should be `$country`

dionyziz: "the as much" should be "as many"

tests/models/country.php
dionyziz: `Country` should be capitalized in all 3 lines.

tests/models/country.php
vitsalis: Country::findall() or $country::findall()

tests/models/country.php
```diff
@@ -0,0 +1,42 @@
...
+    class CountryTest extends UnitTest {
+        protected $countries = array(
+            'GR' => 'Greece',
+            'UK' => 'England',
+            'RU' => 'Russia'
+        );
+        
+        protected function insertCountries() {
+            $countries = $this->countries;
+            $shortnames = array_keys( $countries );
```
vitsalis: you don't use the $shortnames array

tests/models/country.php
```diff
@@ -0,0 +1,42 @@
...
+            $names = array_values( $countries );
```
vitsalis: you don't use the $names array

tests/models/country.php
```diff
@@ -0,0 +1,41 @@
...
+                $country = new Country();
+                $country->shortname = $shortname;
+                $country->name = $name;
+                $country->save();
+            }
+        }
+        public function testFindAll() {
+            $this->insertCountries();
+            $countriesArray = Country::findAll(); 
+            $this->assertTrue( is_array( $countriesArray ), 'Country::findall() did not return an array' );
```
vitsalis: did not return an array of what?

tests/models/country.php
```diff
@@ -0,0 +1,41 @@
...
+            $this->assertTrue( count( $countriesArray ) > 3, 'Country::findall() did not return the as many countries as it should' );
```
vitsalis: "the as many" ?

tests/models/country.php
```diff
@@ -0,0 +1,41 @@
...
+            $shortnames = array_keys( $countries );
+            $shortname = $shortnames[ 0 ];
+            $this->insertCountries();
+            $country = new Country( $testId );
+            $this->assertEquals( $testId, $country->id, "'new Country()' did not return the appropriate country" );
+            $this->assertEquals( $country->shortname, $shortname, "'new Country()' did not return the appropriate shortname" );
+            $this->assertEquals( $country->name, $countries[ $shortname ], "'new Country()' did not return the appropriate name" );
+        }
+    }
+
```
vitsalis, line 49: maybe add "testFindNonExistentCountry()" ?

tests/models/country.php
dionyziz: change `findall` to `findAll`, and similarly below

tests/models/country.php
```diff
@@ -0,0 +1,56 @@
...
+            catch ( ModelNotFoundException $e ) {
+                $caught = true;
+            }
+            $this->assertTrue( $caught, 'When we try to find a non existent country we must get a ModelNotFoundException' );
+        }
+        public function testFindAll() {
+            $this->insertCountries();
+            $countriesArray = Country::findAll(); 
+            $this->assertTrue( is_array( $countriesArray ), 'Country::findAll() did not return an array of countries' );
+            $this->assertTrue( count( $countriesArray ) == 3, 'Country::findAll() did not return as many countries as it should' );
```
vitsalis:
```
===
```

tests/models/country.php
```diff
@@ -0,0 +1,56 @@
...
+            $countries = $this->countries;
+            $testId = 1;
+            $shortnames = array_keys( $countries );
+            $shortname = $shortnames[ 0 ];
+            $this->insertCountries();
+            $country = new Country( $testId );
+            $this->assertTrue( is_object( $country ), "'new Country()' did not return an object" );
+            $this->assertEquals( $testId, $country->id, "'new Country()' did not return the appropriate country" );
+            $this->assertEquals( $country->shortname, $shortname, "'new Country()' did not return the appropriate shortname" );
+            $this->assertEquals( $country->name, $countries[ $shortname ], "'new Country()' did not return the appropriate name" );
```
vitsalis: expected first, then actual

tests/models/country.php
```diff
@@ -0,0 +1,56 @@
...
+        public function testCountryConstruct() {
+            $countries = $this->countries;
+            $testId = 1;
+            $shortnames = array_keys( $countries );
+            $shortname = $shortnames[ 0 ];
+            $this->insertCountries();
+            $country = new Country( $testId );
+            $this->assertTrue( is_object( $country ), "'new Country()' did not return an object" );
+            $this->assertEquals( $testId, $country->id, "'new Country()' did not return the appropriate country" );
+            $this->assertEquals( $country->shortname, $shortname, "'new Country()' did not return the appropriate shortname" );
```
vitsalis: expected first, then actual

PR #29, opened 2014-02-04
Ascii salt

dionyziz:


models/test/withfixtures.php
```diff
@@ -0,0 +1,21 @@
...
+        protected function buildUser( $username ) {
+            $user = new User();
+            $user->username = $username;
+            $user->password = 'secret1234';
+            $user->email = "$username@gmail.com";
+            $user->save();
+
+            return $user;
+        }
+        protected function buildCountry( $name, $shortname ) {
```
pkakelas, line 12: We could rename this function to buildDefaultCountry and make it create a default country (for example Greece).

vitsalis, line 12: I think it would be better for us to just know that buildCountry makes a country with $name and $shortname.

dionyziz, line 12: What are the benefits of renaming it?

models/user.php
```diff
@@ -86,6 +86,25 @@ protected function validate() {
...
+            }
+            else {
+                $this->countryid = 0;
+            }
+            if ( isset( $this->image ) ) {
+                $this->imageid = $this->image->id;
+            }
+            else {
+                $this->imageid = 0;
+            }
```
vitsalis, line 107: why move these to validate? you're not validating anything here.
are you doing this for DRY?

PR #31, opened 2014-02-04
Add some more user tests

vitsalis:


tests/models/user.php
```diff
@@ -45,6 +45,65 @@ public function testSetCountry() {
...
             $this->assertEquals( 1, $user->country->id, 'Country must be the one associated during update' );
         }
+        public function testDuplicateUsername() {
+            $cought = false;
+            $user1 = $this->buildUser( 'pkakelas' );
+            try {
+                $user2 = $this->buildUser( 'pkakelas' );
+            }
+            catch ( ModelValidationException $e ) {
+                $cought = true;
```
dionyziz: `$caught` and similarly below

tests/models/user.php
```diff
@@ -45,6 +45,65 @@ public function testSetCountry() {
...
+                $this->assertEquals(
+                    'username_used',
+                    $e->error,
+                    "If the username is used we must get an 'username_used' error"
```
dionyziz: `an` -> `a`

tests/models/user.php
```diff
@@ -45,6 +45,65 @@ public function testSetCountry() {
...
+                );
+            }
+            $this->assertTrue( $cought, 'A ModelValidationException must be cought if we try to make a duplicate username' );
```
dionyziz: `caught`

PR #37, opened 2014-02-05
Fix a user bug and add testing for it

vitsalis:


models/user.php
```diff
@@ -119,6 +119,11 @@ protected function onBeforeCreate() {
             $this->generateSessionId();
         }
 
+        protected function onCreate() {
```
dionyziz: Shouldn't this also run onUpdate? Suggest:
- Rename current onSave() method to onBeforeSave()
- Create a new onSave() method that runs after saving
- Use onSave() here, so that it also runs with updates

tests/models/user.php
```diff
@@ -124,6 +123,13 @@ public function testJsonSerialize() {
             $this->assertTrue( isset( $data->userid ), 'userid must exist in exported JSON' ); 
             $this->assertEquals( $user->id, $data->userid, 'userid must be encoded properly to JSON' );
         }
+        public function testAuthenticationAfterRenewSessionId() {
```
dionyziz, line 126: nice one :)

models/user.php
vitsalis: no this should run after creation.
The problem was that we may have:

```
$user = new User();
$user->username = 'dionyziz';
$user->password = 'secret1234';
$user->email = 'dionyziz@gmail.com';
$user->save(); // if we don't unset $user->password and $user->salt after save, we will have a problem on $user->renewSessionId
$user->renewSessionId(); // user->update() runs, if password isset it encodes it
```

PR #41, opened 2014-02-07
Add findByEmail() and fix email_used bug

pkakelas:


models/user.php
```diff
@@ -130,8 +140,10 @@ protected function onCreateError( $e ) {
                 throw new ModelValidationException( 'username_used' );
             }
             catch ( ModelNotFoundException $e ) {
+                $other_user = User::findByEmail( $this->email );
```
dionyziz: This will throw a ModelNotFoundException to the controller instead of a ModelValidationException( 'email_used' ) when e-mail is used, so it is wrong. Write a regression test for this?

models/user.php
```diff
@@ -15,7 +15,7 @@ class User extends ActiveRecordBase {
         public $boturl;
         protected $dob;
         protected $tableName = 'users';
-
+
```
dionyziz: Avoid trailing whitespace.

models/user.php
```diff
@@ -130,8 +140,11 @@ protected function onCreateError( $e ) {
                 throw new ModelValidationException( 'username_used' );
             }
             catch ( ModelNotFoundException $e ) {
-                throw new ModelValidationException( 'email_used' );
+                if ( $other_user = User::findByEmail( $this->email ) ) {
```
dionyziz: Just check to see if findByEmail was successful or if it threw an exception, don't compare with anything.

models/user.php
```diff
@@ -130,8 +140,11 @@ protected function onCreateError( $e ) {
...
+                    throw new ModelValidationException( 'email_used' );
+                }
             }
+            throw new DBException( mysql_error() );
```
dionyziz: You already have a DBException, use it.

pkakelas: How should I use it ?

dionyziz: throw $e;

PR #50, opened 2014-02-22
Next round

vitsalis:


views/game/view.php
```diff
@@ -1 +1,5 @@
-Game created.
+<?php
+    ?><h1>Winner Page</h1>
+    <p>The winner is:</p>
+    <p><?php echo $user->username; ?></p><?php
```
dionyziz: Style: New line after `<?php`. New line before `?>`.

models/user.php
```diff
@@ -47,11 +47,12 @@ public function __construct( $id = false ) {
             if ( $id ) {
                 // existing active record object
                 try {
-                    $user_info = dbSelectOne( 'users', [ 'dob', 'username', 'email', 'countryid', 'avatarid' ], compact( "id" ) );
+                    $user_info = dbSelectOne( 'users', [ 'boturl', 'dob', 'username', 'email', 'countryid', 'avatarid' ], compact( "id" ) );
                 }
                 catch ( DBException $e ) {
                     throw new ModelNotFoundException();
                 }
+                $this->boturl = $user_info[ 'boturl' ];
```
dionyziz, line 69: Regression test

models/round.php
```diff
@@ -43,7 +43,7 @@ public function __construct( $a = false, $b = false ) {
                     $user_info = dbSelectOne(
                         'creatures',
                         [ 'userid' ],
-                        compact( 'id' )
+                        compact( 'id', 'gameid' )
```
dionyziz, line 46: Why do you need `gameid`?

models/grader/grader.php
```diff
@@ -113,6 +129,17 @@ public function nextRound() {
                     $error->save();
                 }
             }
+
+            go( 'game', 'update', [ 'gameid' => $this->game->id ] );
```
dionyziz: Don't use `go` in models, only in controllers.

models/grader/grader.php
```diff
@@ -86,6 +91,17 @@ public function nextRound() {
...
+            foreach ( $round->creatures as $creature ) {
+                if ( $creature->alive ) {
+                    $usersAlive[ $creature->user->id ] = $creature->user;
+                }
+            }
+
+            if ( count( $usersAlive ) === 1 ) {
+                throw new WinnerException( end( $usersAlive )->id );
+            }
+
```
dionyziz, line 104: test

models/grader/grader.php
```diff
@@ -21,6 +21,11 @@ public function __construct( Game $game, $users = false ) {
                 assert( $users === false, 'game already has users since genesis has run' );
                 assert( isset( $game->users ), 'game must have users' );
                 $this->registeredUsers = $game->users;
+                foreach ( $game->users as $user ) {
+                    $bot = new GraderBot( $user );
+                    $bot->game = $game;
+                    $this->registeredBots[] = $bot;
+                }
```
dionyziz, line 28: regression test

models/game.php
```diff
@@ -76,7 +76,7 @@ public function genesis() {
             $this->rounds[ 0 ] = new Round();
             $this->rounds[ 0 ]->game = $this;
             $this->rounds[ 0 ]->id = 0;
-            $id = 0;
+            $id = 1;
```
dionyziz, line 79: Regression test

models/game.php
```diff
@@ -100,6 +101,7 @@ public function genesis() {
                     $this->rounds[ 0 ]->creatures[] = $creature;
                 }
             }
+            $this->rounds[ 0 ]->save();
```
dionyziz, line 104: regression test

models/round.php
vitsalis, line 46: There are a lot of creatures with the same id. But there are not creatures with the same id and gameid.
Each creature id starts from 1, in each game remember?

models/game.php
```diff
@@ -87,6 +87,7 @@ public function genesis() {
                     $creature->hp = $this->maxHp;
                     $creature->alive = true;
                     $creature->intent = new Intent( ACTION_NONE, DIRECTION_NONE );
+                    $creature->save();
```
dionyziz, line 90: regression test

models/round.php
dionyziz, line 46: Regression test this.

models/user.php
vitsalis, line 69: In order to test this we need to add update boturl on user update. I suggest we live this be as it is for now, and test/add the feature on another branch

dionyziz, line 69: You don't need to add it on update; just create a user and look them up.

header.php
```diff
@@ -6,4 +6,5 @@
     ini_set( 'display_errors', '1' );
     date_default_timezone_set( "Europe/Athens" );
     require_once 'models/assert.php';
+    ini_set('max_execution_time', 60);
```
dionyziz: style

models/grader/grader.php
```diff
@@ -113,6 +129,24 @@ public function nextRound() {
                     $error->save();
                 }
             }
+
+            throw new NextRoundException( $this->game->id );
```
dionyziz: Exceptions should be thrown, as the name suggests, during exceptional cases. This exception is not needed - exceptions must be used only to break execution flow, usually under error conditions, until caught. If the next round is executed successfully as the game progresses, you can just return normally.

A winner exception is OK, as winning a game is unusual - when we call nextRound, we expect a new round to be created and the game to advance, but when a winner is detected, the game is over, which is an exceptional situation.

However, it is usually best to limit exceptions to erroneous conditions. I suggest that we change this function to return a value based on whether a winner exists or not.

The NextRoundException must definitely be avoided, however.

tests/models/grader/grader.php
```diff
@@ -241,6 +249,41 @@ public function testRoundResolution() {
...
+            $game->rounds[ 0 ]->creatures = [ 1 => $this->buildCreature( 1, 1, 1, $game->users[ 1 ] ) ];
+
+            return $game;
+        }
+        public function testFindBotsFromGame() {
+            $game = $this->buildGameWithUserAndCreature();
+            $grader = new Grader( $game );
+
+            $this->assertTrue( isset( $grader->registeredUsers ), "Grader must get it's users from the game" );
+            $this->assertTrue( isset( $grader->registeredBots ), "Grader must get it's bots from the game" );
```
dionyziz: :s/it's/its

models/round.php
```diff
@@ -33,18 +33,28 @@ public function __construct( $a = false, $b = false ) {
...
-                    compact( 'roundid', 'gameid' )
-                );
+                try {
+                    $creatures_info = dbSelect(
+                        'roundcreatures',
+                        [ 'creatureid', 'action', 'direction', 'hp', 'locationx', 'locationy' ],
+                        compact( 'roundid', 'gameid' )
+                    );
+                }
+                catch ( DBException $e ) {
```
dionyziz: Why is this needed?

models/round.php
```diff
@@ -33,18 +33,28 @@ public function __construct( $a = false, $b = false ) {
...
+                }
                 foreach ( $creatures_info as $i => $creature_info ) {
                     $id = $creature_info[ 'creatureid' ];
-                    $user_info = dbSelectOne(
-                        'creatures',
-                        [ 'userid' ],
-                        compact( 'id' )
-                    );
+                    try {
+                        $user_info = dbSelectOne(
```
dionyziz: Why `dbSelectOne`?

vitsalis: ids are unique for each game. So id and game have to be unique on a table. Also a creature belongs to one user.

dionyziz: OK. Why do we need the try/catch?

dionyziz:
Please merge from master.

models/grader/grader.php
```diff
@@ -113,6 +129,17 @@ public function nextRound() {
                     $error->save();
                 }
             }
+
+            return $this->game->id;
```
dionyziz: Why does this function return the game id? We already know the game id from `$grader->game->id`. It doesn't need to return anything.

tests/models/grader/grader.php
```diff
@@ -155,7 +155,11 @@ protected function nextRoundAndGetErrors( $responseValid ) {
             }
             $grader->initiate();
             $grader->createGame();
-            $grader->nextRound();
+            try {
+                $grader->nextRound();
+            }
+            catch ( NextRoundException $e ) {
```
dionyziz: Fix your tests. Similarly bellow.

dionyziz:
This looks good overall. Please fix the things I indicated and I will merge.

tests/models/grader/grader.php
vitsalis: my tests work.

dionyziz: `NextRoundException` is not defined. How can they possibly be working?

models/creature.php
```diff
@@ -15,8 +15,9 @@ class Creature extends ActiveRecordBase {
         protected static $attributes = [ 'id', 'gameid', 'userid' ];
         protected static $tableName = 'creatures';
 
-        public function __construct( $creature_info = [] ) {
-            if ( !empty( $creature_info ) ) {
+        public function __construct( $a = false, $b = false, $c = false ) {
+            if ( is_array( $a ) ) {
+                $creature_info = $a;
```
dionyziz, line 20: This way of constructing an active record object mapping to an existing database record by passing it its own information through a dictionary is good. We should generalize this to other models for optimization reasons after we benchmark. Please keep this in mind as we move forward. Let's discuss this for the next patch.

PR #99, opened 2014-02-24
Move update on active record base

vitsalis:


models/base.php
```diff
@@ -34,6 +42,25 @@ protected function create() {
             $this->onCreate();
         }
 
+        protected function update() {
```
dionyziz, line 45: Excellent abstraction, good job.

dionyziz, line 45: Unit test this? We have limited unit tests for ActiveRecordBase. As it's one of our base classes, I feel we need to test it more extensively, including this `update`.

PR #101, opened 2014-02-27
Front end functions

vitsalis:


models/db.php
```diff
@@ -60,14 +60,20 @@ function dbDelete( $table, $where ) {
         return mysql_affected_rows();
     }
 
-    function dbSelect( $table, $select = [ "*" ], $where = [] ) {
+    function dbSelect( $table, $select = [ "*" ], $where = [], $orderBy = false, $limit = false ) {
```
dionyziz, line 81: Please test the new `orderBy` and `limit` arguments. Generally, I feel our db model is undertested. I recommend that we add more tests for all functions in the long run, because it is used extensively throughout the codebase. For now, please test the things you edit/add.

Feel free to create a github issue to write more tests for the db model and assign to me.

dionyziz, line 81: Assert that `$limit` is a positive number.

models/game.php
```diff
@@ -19,6 +19,23 @@ class Game extends ActiveRecordBase {
         protected static $tableName = 'games';
         protected static $attributes = [ 'width', 'height', 'created' ];
 
+        public static function findByDatetime( $datetime ) {
```
dionyziz: This is not good. The front-end can't know the exact second a game took place. Make it so that it finds the next game after a given date.

models/game.php
```diff
@@ -19,6 +19,23 @@ class Game extends ActiveRecordBase {
...
+                $game = dbSelectOne( 'games', [ 'id' ], compact( 'created' ) );
+            }
+            catch ( DBException $e ) {
+                throw new ModelNotFoundException();
+            }
+            return new Game( $game[ 'id' ] );
+        }
+
+        public static function getLastGame() {
+            $game = dbSelect( 'games', [ 'id' ], [], 'created ASC', 1 );
```
dionyziz: This will return the first game. Please fix, then regression test.

models/game.php
```diff
@@ -58,16 +75,40 @@ public function initiateAttributes() {
...
+                $newUsers = [];
+                foreach ( $this->rounds[ $i ]->creatures as $creature ) {
+                    if ( $creature->alive && !isset( $found[ $creature->user->id ] ) ) {
+                        $newUsers[] = $creature->user;
+                        $found[ $creature->user->id ] = true;
+                    }
+                }
+                if ( !empty( $newUsers ) ) {
+                    ++$position;
+                    $ratings[ $position ] = $newUsers;
```
dionyziz, line 82: This is incorrect. Check the spec for how rankings are defined. Then test based on spec. If there's a draw in the game between two winners, the person to die last before the two winners died should be ranked 3, not 2.

models/game.php
```diff
@@ -58,16 +75,40 @@ public function initiateAttributes() {
...
+                }
+            }
+            return $ratings;
+        }
+
+        public function getCountryRatings( $country ) {
```
dionyziz: type hint

tests/models/game.php
```diff
@@ -146,6 +146,83 @@ public function testGameIdNonZero() {
...
+            $round2->id = 1;
+            $round2->creatures = [ 1 => $creature1, 2 => $creature2Clone, 3 => $creature3Clone ];
+
+            $game = new Game();
+            $game->users = [ 1 => $user1, 2 => $user2, 3 => $user3 ];
+            $game->rounds = [ 0 => $round1, 1 => $round2 ];
+
+            return $game;
+        }
+        public function testGetGlobalRatings() {
```
dionyziz, line 178: This test is good. Write another test testing the case where we have a draw and a third player. The ranking should then be 1 for the first two, 3 for the last player.

tests/models/game.php
```diff
@@ -146,6 +146,83 @@ public function testGameIdNonZero() {
...
+
+            $ratings = $game->getCountryRatings( $country1 );
+
+            $this->assertEquals( 1, count( $ratings[ 1 ] ), 'Only one player must be in the first place' );
+            $this->assertEquals( 1, count( $ratings[ 2 ] ), 'All the players that were defeated on the last round must go to the second place' );
+
+            $this->assertEquals( $game->users[ 1 ]->id, $ratings[ 1 ][ 0 ]->id, 'The ratings must contain the valid players' );
+            $this->assertEquals( $game->users[ 3 ]->id, $ratings[ 2 ][ 0 ]->id, 'The ratings must contain the valid players' );
+        }
+        public function testGetLastGame() {
```
dionyziz, line 212: Make this test distinguish between two games: The first and the last.

models/db.php
vitsalis, line 81: I think testing about db must be done on another branch. I suggest we just let it be for now and test in another branch.

dionyziz, line 81: OK

models/game.php
```diff
@@ -52,22 +58,49 @@ public function __construct( $id = false ) {
...
+                foreach ( $this->rounds[ $i ]->creatures as $creature ) {
+                    if ( $creature->alive && !isset( $found[ $creature->user->id ] ) ) {
+                        $newUsers[] = $creature->user;
+                        $found[ $creature->user->id ] = true;
+                    }
+                }
+                if ( !empty( $newUsers ) ) {
+                    $position += 1;
+                    $ratings[ $position ] = $newUsers;
+                    $position += count( $newUsers ) - 1;
```
dionyziz: For simplicity, recommend that you remove the `- 1` from here, remove the `+= 1` line completely, and start `$position` from 1.

vitsalis: that was really smart, thanks

models/db.php
```diff
@@ -78,14 +78,14 @@ function dbDelete( $table, Array $where = [] ) {
...
         if ( empty( $where ) ) {
-            return dbSelectMulti( $table, $select );
+            return dbSelectMulti( $table, $select, [], $orderBy, $limit );
         }
         return dbSelectMulti( $table, $select, [ $where ] );
```
dionyziz: Why are `$orderBy` and `$limit` not passed here? Suggest regression test.

models/db.php
```diff
@@ -78,14 +78,14 @@ function dbDelete( $table, Array $where = [] ) {
         return mysql_affected_rows();
     }
 
-    function dbSelect( $table, $select = [ "*" ], $where = [] ) {
+    function dbSelect( $table, $select = [ "*" ], $where = [], $orderBy = false, $limit = false ) {
         if ( empty( $where ) ) {
-            return dbSelectMulti( $table, $select );
+            return dbSelectMulti( $table, $select, [], $orderBy, $limit );
```
dionyziz, line 83: This seems redundant; the line below can handle the empty where case.

vitsalis, line 83: it doesn't work if I don't do this. the query becomes
SELECT .. FROM .. WHERE () IN (())
even if it becomes SELECT ... FROM ... WHERE () IN () it doesn't work

PR #108, opened 2014-03-01
Travis

dionyziz:


dionyziz:
Fixes #57

controllers/testrun.php
```diff
@@ -21,10 +24,20 @@ public function create( $name, $all = false ) {
...
+                        $failed = true;
+                    }
+                }
 
                 $unittests[] = $unittest;
             }
             require_once 'views/testrun/results.php';
+
+            if ( $failed ) {
+                return 1;
```
pkakelas, line 38: You should use true;

controllers/testrun.php
```diff
@@ -21,10 +24,20 @@ public function create( $name, $all = false ) {
...
+            }
+            return 0;
```
pkakelas, line 40: false;

views/testrun/results.txt.php
```diff
@@ -0,0 +1,32 @@
+<?php
+    echo "\n\n";
+    echo "================\n";
+    echo pluralize( count( $unittests ), 'unit test' ) . "\n";
+    echo "================\n\n";
+?>
```
pkakelas: Why do you close the ?> and in the next line you open it again?

controllers/testrun.php
dionyziz, line 40: No - this is a shell [exit status](http://tldp.org/LDP/abs/html/exit-status.html) and must be a byte numeral (from 0 to 255).

pkakelas, line 40: Ok then.

dionyziz:
CLI for test failure:

image

dionyziz:
CLI for test success:

image

PR #123, opened 2014-03-06
Ranking page

pkakelas:
It is ugly
screenshot from 2014-03-06 22 52 42

### DIFF
--- a/views/home.php
+++ b/views/home.php
@@ -3,24 +3,26 @@
 ?>
 <div id="home">
     <p class='center'>Last match ratings(global)</p>
-    <ol class='center' id="toplist">
+    <ul class='center' id="toplist">
         <?php
+            $i = 1;
             foreach( $ratings as $rating ) {
                 foreach ( $rating as $user ) {
-                    ?><li><a href=""><?php echo $user->username; ?></a></li><?php
+                    ?><li><a href=""><?php echo "$i.$user->username"; ?></a></li><?php
                 }
+                $i++;
             } 
         ?>
-    </ol>
+    </ul>
     <?php
-//        $filter = [ 
-//            [ 'value' => 'default', 'content' => 'select Ratings' ], 
-//            [ 'value' => 'global', 'content' => 'Global' ], 
-//            [ 'value' => 'country', 'content' => 'By country' ]
-//        ];
-//        $form = new Form( 'dashboard', 'view' );
-//        $form->id = 'selectbox';
-//        $form->createSelect( 'filter', '', $filter );
+//      $filter = [ 
+//          [ 'value' => 'default', 'content' => 'select Ratings' ], 
+//          [ 'value' => 'global', 'content' => 'Global' ], 
+//          [ 'value' => 'country', 'content' => 'By country' ]
+//      ];
+//      $form = new Form( 'dashboard', 'view' );
+//      $form->id = 'selectbox';
+//      $form->createSelect( 'filter', '', $filter );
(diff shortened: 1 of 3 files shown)


views/home.php
```diff
@@ -1,34 +1,33 @@
...
-            <p>Check your program's process <a href="">here</a></p>
-            <?php
-                if ( isset( $user ) ) {
-                    require_once 'views/session/logoutform.php';
+<?php
+    require_once 'views/header.php';
+?>
+<div id="home">
+    <p class='center'>Last match ratings(global)</p>
+    <ol class='center' id="toplist">
```
dionyziz, line 6: This should be `ul` because it is a rank. If you want to hide the numbers, use CSS.

views/home.php
```diff
@@ -1,34 +1,33 @@
...
+        <?php
+            $i = 1;
+            foreach( $ratings as $rating ) {
+                foreach ( $rating as $user ) {
+                    ?><li><a href=""><?php echo "$i.$user->username"; ?></a></li><?php
                 }
-            ?>
-        </div>
-    </div>
+                $i++;
```
dionyziz, line 13: Why do you need `$i`?

views/home.php
```diff
@@ -1,34 +1,33 @@
...
-            <p>Check your progress <a href="">here</a></p>
-            <p>Check your program's process <a href="">here</a></p>
-            <?php
-                if ( isset( $user ) ) {
-                    require_once 'views/session/logoutform.php';
+<?php
+    require_once 'views/header.php';
+?>
+<div id="home">
+    <p class='center'>Last match ratings(global)</p>
```
dionyziz, line 5: Where is class `center` defined?

views/home.php
```diff
@@ -1,34 +1,33 @@
...
+//      $filter = [
+//          [ 'value' => 'default', 'content' => 'select Ratings' ],
+//          [ 'value' => 'global', 'content' => 'Global' ],
+//          [ 'value' => 'country', 'content' => 'By country' ]
+//      ];
+//      $form = new Form( 'dashboard', 'view' );
+//      $form->id = 'selectbox';
+//      $form->createSelect( 'filter', '', $filter );
+        if ( isset( $user ) ) {
+            require_once 'views/session/logoutform.php';
```
dionyziz, line 27: This should be on the header.

views/home.php
pkakelas, line 6: Yes, but we want to set the "order" of the last game's best players. This
is why I prefer ordered list that unordered
Óôéò 7 Ìáñ 2014 12:35 ð.ì., ï ÷ñÞóôçò "Dionysis Zindros" <
notifications@github.com> Ýãñáøå:

views/home.php
pkakelas, line 13: More than 1 players may be sharing a position. In this situation Vit
suggested me to do something like that.
1.Vitsalis
2.Dionyziz
2.pkakelas
3.gtklocker
So I use this counter to replace the default number in the ordered list
with a custom according to the situation. If you have a better idea pls
suggest.
Óôéò 7 Ìáñ 2014 12:36 ð.ì., ï ÷ñÞóôçò "Dionysis Zindros" <
notifications@github.com> Ýãñáøå:

dionyziz, line 13: OK fine leave it like this for now.

views/home.php
dionyziz, line 6: OK leave it as is for now.

PR #137, opened 2014-03-10
Rename errors.error to errors.description on database

vitsalis:


dionyziz:
Please test that errors that occur in nextRound are saved and loaded correctly from the database.

Issue #72, opened 2014-02-22
Use sane defaults in config.php


dionyziz:

dionyziz:
config-local.php should only need to specify the keys it wants to make different from the default config.php file. config.php should have some reasonable defaults. The different keys in config-local.php should then be merged into config.php when config-local is loaded. config-local can look like this:

```
return [
    'development' => [
        'db' => [
            'username' => 'endofcodes',
            'password' => 'secret'
        ]
    ]
]
```

This means that the 'testing' environment will be completely copied from config.php, every section outside of 'db' will be identical to 'config' (e.g. `$config[ 'base' ]` will be set from config.php), and, in the db section, only the username and password keys will be overridden, meaning for example that the `host` key will remain the same.

PR #140, opened 2014-03-11
Use sane defaults on config.php

vitsalis:


vitsalis:
Fixes #72

config/config.php
```diff
@@ -1,69 +1,66 @@
...
-            ]
-        ];
-        return $config;
-    }
+            'files' => [
+                'avatar_extentions' => [ 'jpg', 'png', 'jpeg' ],
+                'flag_extention' => 'png'
+            ],
+            'paths' => [
+                'avatar_path' => 'sample_path',
```
dionyziz: Use a better default

config/config.php
```diff
@@ -1,69 +1,66 @@
...
+                'user' => 'travis',
+                'pass' => '',
+                'dbname' => 'endofcodes_test'
+            ],
+            'files' => [
+                'avatar_extentions' => [ 'jpg', 'png', 'jpeg' ],
+                'flag_extention' => 'png'
+            ],
+            'paths' => [
+                'avatar_path' => 'sample_path',
```
dionyziz: Use a better defaut; "avatars"?

config/config.php
```diff
@@ -1,69 +1,66 @@
...
-                'email' => 'team@endofcodes.com',
-                'base' => 'http://localhost/endofcodes/',
-                'root' => '/var/www/endofcodes/'
+    return [
+        'development' => [
+            'db' => [
+                'host' => 'sample_host',
+                'user' => 'sample_user',
+                'pass' => 'sample_pass',
+                'dbname' => 'sample_dbname'
```
dionyziz: Use sane defaults here. Some sane defaults are 'localhost' for host, 'endofcodes' for user and dbname. Password of course can't have a sane default, so this is fine.

PR #152, opened 2014-03-18
Fixed footer bug

pkakelas:
Βug...
screenshot from 2014-03-19 14 57 51
Fixed.
screenshot from 2014-03-19 14 59 48

### DIFF
--- a/static/style/general.css
+++ b/static/style/general.css
@@ -10,5 +10,5 @@
     margin-left: 46%;
 }
 #main {
-    margin-bottom: 5%;
+    margin-bottom: 65px;
 }
--- a/static/style/general.css
+++ b/static/style/general.css
@@ -9,3 +9,6 @@
 #avatar-form {
     margin-left: 46%;
 }
+#main {
+    margin-bottom: 5%;
+}
--- a/static/style/general.css
+++ b/static/style/general.css
@@ -9,6 +9,3 @@
 #avatar-form {
     margin-left: 46%;
 }
-#main {
-    margin-bottom: 65px;
-}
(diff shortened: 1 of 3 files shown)


dionyziz:
Please upload screenshots for before and after the bugfix.

static/style/general.css
```diff
@@ -9,3 +9,6 @@
 #avatar-form {
     margin-left: 46%;
 }
+#main {
+    margin-bottom: 5%;
```
dionyziz: I think you want to use an absolute value for this. Isn't the footer static in height regardless of window size? This means you need to make two changes - one here for the margin, and one for the actual size of the footer height.

static/style/general.css
```diff
@@ -9,3 +9,6 @@
 #avatar-form {
     margin-left: 46%;
 }
+#main {
+    margin-bottom: 65px;
+}
```
dionyziz: Move to same file that determines footer height to ensure orthogonality between files.

dionyziz:
I don't think `position: static` will work well with pages of short height. Can you post a screenshot with a page that doesn't fill the screen vertically?

static/style/general.css
dionyziz: I reiterate this comment.

pkakelas: I don't believe that it is appropriate to move this code to footer.css
becsuse this property is about the main div of the page and not the footer
Óôéò 22 Ìáñ 2014 1:59 ð.ì., ï ÷ñÞóôçò "Dionysis Zindros" <
notifications@github.com> Ýãñáøå:

dionyziz: The property of `margin-bottom` concerns the footer of the page. `main` is making space for the footer. If the footer changes in height, main's margin should also change. We should keep these changes in the same file.

PR #159, opened 2014-03-20
Add utf8

vitsalis:


views/header.php
```diff
@@ -3,6 +3,7 @@
 <html lang="en">
     <head>
         <title>EndofCodes Demo</title>
+        <meta charset="utf-8">
```
dionyziz: Please maintain XML validity by closing all HTML tags.

dionyziz:
Before/after screenshots please.

views/header.php
```diff
@@ -3,6 +3,7 @@
 <html lang="en">
     <head>
         <title>EndofCodes Demo</title>
+        <meta charset="utf-8" />
```
dionyziz, line 5: Let's move this above the title, just in case we ever want to have UTF-8 characters in the title.

PR #160, opened 2014-03-20
Remove description from assert

vitsalis:


models/db.php
```diff
@@ -110,7 +110,7 @@ function dbSelectMulti( $table, $select = [ "*" ], $wheres = [], $orderBy = fals
             $sql .= ' ORDER BY ' . $orderBy;
         }
         if ( $limit !== false ) {
-            assert( $limit > 0, 'limit must be a positive integer' );
+            assert( $limit > 0 );
```
dionyziz: I think we should keep the descriptions as comments.

PR #188, opened 2014-03-28
Graceful messages when db credentials are wrong

vitsalis:


vitsalis:
Fixes #62

models/database.php
```diff
@@ -1,9 +1,12 @@
 <?php
+    function throwDBException( $message ) {
+        throw new DBException( $message );
+    }
     function dbInit() {
         global $config;
 
-        mysql_connect( $config[ 'db' ][ 'host' ], $config[ 'db' ][ 'user' ], $config[ 'db' ][ 'pass' ] ) or die( 'Failed to connect to MySQL: ' . mysql_error() );
+        mysql_connect( $config[ 'db' ][ 'host' ], $config[ 'db' ][ 'user' ], $config[ 'db' ][ 'pass' ] ) or throwDBException( 'Failed to connect to MySQL: ' . mysql_error() );
```
pkakelas, line 8: I believe we should have a different exception for situations like wrong credentials e.t.c.

vitsalis, line 8: `mysql_error()` covers that I think

models/database.php
```diff
@@ -1,9 +1,12 @@
...
 
-        mysql_select_db( $config[ 'db' ][ 'dbname' ] ) or die( 'Failed to select MySQL database: ' . mysql_error() );
+        mysql_select_db( $config[ 'db' ][ 'dbname' ] ) or throwDBException( 'Failed to select MySQL database: ' . mysql_error() );
```
dionyziz, line 10: Why don't you just `throw` and need a function?

PR #212, opened 2014-04-08
Migrate population script to active record and merge population and migration scripts

pkakelas:


database/migration/20140408205938_populate_country.php
```diff
@@ -0,0 +1,11 @@
+<?php
```
dionyziz, line 1: This should follow the usual migration OOP API.

database/migration/20140408205938_populate_country.php
```diff
@@ -0,0 +1,11 @@
...
+    require_once 'models/country.php';
+    require_once 'population/countries_array.php';
```
dionyziz: Rename `population` folder to `data`?

database/migration/20140408205938_populate_country.php
```diff
@@ -0,0 +1,11 @@
...
+
+    $countries = getCountries();
+
+    foreach ( $countries as $countryShortname => $countryName ) {
+        $country = new Country( '', $countryName, $countryShortname );
```
dionyziz: Please follow standard active record patterns; constructor should not take values like this. Attributes should be defined by setting their values on the object:

```
$country->name = $countryName
```

models/country.php
```diff
@@ -31,6 +31,10 @@ public function __construct( $id = false ) {
                 $this->shortname = $row[ 'shortname' ];
                 $this->flag = $config[ 'paths' ][ 'flag_path' ] . strtolower( $this->shortname ) . '.' . $config[ 'files' ][ 'flag_extention' ];
             }
+            else {
+                $this->name = $name;
+                $this->shortname = $shortname;
+            }
```
dionyziz: All changes to this file are completely unnecessary.

PR #226, opened 2014-04-18
Prevent "file not found" warnings on Migrationrun

### LINKED ISSUE #222: File not found on Migrationrun
untitled
When you try to run the migrations for the first time on a fresh installation.

bugos:
Fixes #222.
I am not sure if supressing the warnings is what we want. Please review.


dionyziz:
You should check for file existence instead of suppressing warnings. `@` is generally bad practice.

models/migration.php
```diff
@@ -47,9 +47,10 @@ protected static function getUnexecuted( $env ) {
         }
 
         public static function findLast( $env = '' ) {
-            if ( !$logs = file_get_contents( static::$log ) ) {
+            if ( !file_exists( static::$log ) ) {
```
pkakelas, line 51: Why don't you just create an empty log file? In this way, if the developer haven't used the migrationrun so far, he will get an error.

bugos, line 51: Yes, you are right :)

models/migration.php
```diff
@@ -47,7 +47,8 @@ protected static function getUnexecuted( $env ) {
         }
 
         public static function findLast( $env = '' ) {
-            if ( !$logs = file_get_contents( static::$log ) ) {
+            if ( !file_exists( static::$log ) && !touch( static::$log ) 
+              || !$logs = file_get_contents( static::$log ) ) {
```
dionyziz: Unusual coding paradigm in PHP; suggest actions to be each on a separate line and if to be used only for checking:

```
$touched = touch( static::$log );
if ( !$touched ) {... 
```

models/migration.php
```diff
@@ -47,7 +47,11 @@ protected static function getUnexecuted( $env ) {
         }
 
         public static function findLast( $env = '' ) {
-            if ( !$logs = file_get_contents( static::$log ) ) {
+            $touched = true;
+            if ( !file_exists( static::$log ) ) {
+                $touched = touch( static::$log );
+            }
+            if ( $touched && !$logs = file_get_contents( static::$log ) ) {
```
dionyziz: Since we are making modifications on this, I suggest that the assignment to `$logs` be also put on a separate line. `if` conditions are better kept as checks, not actions. I know the code before the patch was like this, but it's good to fix it while we're at it.

bugos:
I think that we should be more strict with situations when we cannot access the history file. 
If we know that the history file exists but we cannot read the previous migrations from it, we should not let the user run any new migrations at all. 

Also I don't think ModelNotFoundException is appropriate here.

models/migration.php
```diff
@@ -47,7 +47,14 @@ protected static function getUnexecuted( $env ) {
...
-            if ( !$logs = file_get_contents( static::$log ) ) {
+            $touched = true;
+            if ( !file_exists( static::$log ) ) {
+                $touched = touch( static::$log );
+            }
+            if ( $touched ) {
+                $logs = file_get_contents( static::$log );
+            }
+            if ( empty( $logs ) ) {
                 throw new ModelNotFoundException();
```
dionyziz, line 58: `ModelNotFoundException` is appropriate here. Please talk to @VitSalis about migrations and exceptions. Migrations used to be a separate entity from models, but we recently converted migrations to be models, so this is appropriate here: When `findLast` is called and there are no migrations in the history, the last migration model cannot be found, so a `ModelNotFoundException` is thrown correctly.

PR #227, opened 2014-04-20
Remove base from config

### LINKED ISSUE #141: Remove base and root from config
These can be automatically detected so they don't need to be in the config. Figure out a way to detect them and remove them from the config.

bugos:
Fixes #141.
Please test.


helpers/config.php
```diff
@@ -18,6 +18,7 @@ function getConfig( $env ) {
         }
         $config = $config[ $env ];
         $config[ 'root' ] = getcwd();
+        $config[ 'base' ] = dirname( $_SERVER[ 'SCRIPT_NAME' ] ) . '/';
```
dionyziz: In case the script live in the root directory, `dirname` will return `/` and hence this will result in `//`, while the expected output would be just `/`. Furthermore, to adhere to HTTP standards, redirects must feature the full path; in your patch the domain name and protocol are not included, but should.

bugos:
@dionyziz thanks for the review. 
I have updated the pull request. Please confirm that it works as expected.

PR #232, opened 2014-04-29
reformat_countries_to_JSON

pkakelas:
I had tested the migration script many times, I do not get why I had the path mistake. Sorry about that.


dionyziz:
Please use more descriptive commit messages. "init commit" is unacceptable.

PR #235, opened 2014-05-04
Controller was saving the boturl even though it was wrong

vitsalis:


controllers/bot.php
```diff
@@ -11,16 +11,17 @@ public function update( $boturl = '' ) {
             if ( !filter_var( $boturl, FILTER_VALIDATE_URL ) ) {
                 go( 'bot', 'update', [ 'boturl_invalid' => true ] );
             }
-            $user = $_SESSION[ 'user' ];
-            $user->boturl = $boturl; 
-            $bot = new GraderBot( $user );
+            $userclone = clone $_SESSION[ 'user' ];
+            $userclone->boturl = $boturl;
+            $bot = new GraderBot( $userclone );
```
dionyziz: Let's move this check into the model to keep the controller thin.

controllers/bot.php
```diff
@@ -33,6 +30,7 @@ public function updateView( $boturl_empty, $boturl_invalid, $bot_fail, $errorid
                     throw new HTTPUnauthorizedException();
                 }
             }
+            $user = $_SESSION[ 'user' ];
```
dionyziz: Why is this needed now that the `setBoturl` function was moved to the model?

vitsalis: because I use the `$user` object on `views/bot/update.php`

PR #237, opened 2014-05-05
Add dob inputs in update and configured related back end

pkakelas:
screenshot from 2014-05-05 15 15 07

### DIFF
--- a/models/user.php
+++ b/models/user.php
@@ -155,15 +155,13 @@
         }
 
         protected function onBeforeCreate() {
-            $day = intval( $this->dateOfBirth[ 'day' ] );
-            $month = intval( $this->dateOfBirth[ 'month' ] );
-            $year = intval( $this->dateOfBirth[ 'year' ] );
-            if ( !checkdate( $day, $month, $year ) ) {
-                $day = $month = $year = 0;
-            }
-            $dob = $this->dob = $year . '-' . $month . '-' . $day;
             $this->imageid = 0;
             $this->generateSessionId();
+            $dob = $this->dob = $this->prepareDob();
+        }
+
+        protected function onBeforeUpdate() {
+            $dob = $this->dob = $this->prepareDob();
         }
 
         protected function onSave() {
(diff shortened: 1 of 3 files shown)


helpers/html.php
```diff
@@ -222,12 +222,12 @@ function includeStyle( $path ) {
     }
 
     function createSelectPrepare( $array, $first = '', $keys = '' ) {
-        if ( empty( $key ) ) {
+        if ( empty( $keys ) ) {
```
dionyziz: This is a refactoring. I suggest we start writing unit tests for this. It doesn't seem too hard. Let's start with this function only and a regression test for this bug.

models/user.php
```diff
@@ -155,15 +155,13 @@ protected function onBeforeSave() {
...
-            $day = intval( $this->dateOfBirth[ 'day' ] );
-            $month = intval( $this->dateOfBirth[ 'month' ] );
-            $year = intval( $this->dateOfBirth[ 'year' ] );
-            if ( !checkdate( $day, $month, $year ) ) {
-                $day = $month = $year = 0;
-            }
-            $dob = $this->dob = $year . '-' . $month . '-' . $day;
             $this->imageid = 0;
             $this->generateSessionId();
+            $dob = $this->dob = $this->prepareDob();
```
dionyziz: Why do you need the local variable `$dob`?

models/user.php
```diff
@@ -155,15 +155,13 @@ protected function onBeforeSave() {
...
+        }
+
+        protected function onBeforeUpdate() {
+            $dob = $this->dob = $this->prepareDob();
```
dionyziz: Unused local variable `$dob`

models/user.php
```diff
@@ -197,6 +195,16 @@ protected function generateSessionId() {
...
         }
 
+        protected function prepareDob() {
+            $day = intval( $this->dateOfBirth[ 'day' ] );
+            $month = intval( $this->dateOfBirth[ 'month' ] );
+            $year = intval( $this->dateOfBirth[ 'year' ] );
+            if ( !checkdate( $day, $month, $year ) ) {
+                $day = $month = $year = 0;
+            }
+            return $year . '-' . $month . '-' . $day;
```
dionyziz: Suggest simply setting `$this->dob` here since this function is not referentially transparent in its input anyway.

PR #238, opened 2014-05-05
Create select prepare bug and make regression test for it

pkakelas:


helpers/html.php
```diff
@@ -222,12 +222,12 @@ function includeStyle( $path ) {
...
 
     function createSelectPrepare( $array, $first = '', $keys = '' ) {
-        if ( empty( $key ) ) {
+        if ( empty( $keys ) ) {
             $keys = $array;
         }
         $array = array_combine( $keys, $array );
         if ( !empty( $first ) ) {
-            array_unshift( $array, $first );
+            $array = [ 'first' => $first ] + $array;
```
dionyziz: Better written as `$array += compact( 'first' );`

tests/helpers/html.php
```diff
@@ -0,0 +1,31 @@
+<?php
+    class htmlTest extends UnitTestWithFixtures {
```
dionyziz: Follow the naming patterns of the class it tests: HTML in upper-case. Perhaps this can be called more specifically `HTMLHelperTest`? Recommend using specific names for entities that are not models. e.g. we might later have controller tests; in this case we'd name them `UserControllerTest` (VS the model test `UserTest`).

tests/helpers/html.php
```diff
@@ -0,0 +1,31 @@
...
+            $this->assertEquals( $res1[ 'b' ], 'b',  
+                'createPrepareSelect keys must be the same with elements if not any keys are given' );
+            $this->assertEquals( $res2[ 'key1' ], 'a',  
+                'createPrepareSelect keys must be the same with elements if not any keys are given' );
+            $this->assertEquals( $res2[ 'key2' ], 'b',  
+                'createPrepareSelect must combine the array with the keys in order' );
+            $this->assertEquals( $res3[ 'first' ], 'testTitle',  
+                'createPrepareSelect must hold in the key "first" the value given as a second parameter' );
+            $this->assertEquals( array_shift( array_slice( $res3, 0, 1 ) ), 'testTitle',  
+                'createPrepareSelect must return an array with first element the value given as a second parameter' );
```
dionyziz, line 29: Please follow the pattern set-then-assert for each case.

helpers/html.php
pkakelas: I need the $first var to be first in my array

dionyziz: `$array = compact( 'first' ) + $array` ?

dionyziz: Also perhaps this could be called something more insightful than `first`. Suggest `placeholder`, `title`, or `default`.

vitsalis:
Travis failed.

views/user/update.php
```diff
@@ -43,17 +41,6 @@
...
-        $days = createSelectPrepare( range( 1, 31 ), 'Select Day' );
-        $self->createSelect( $days, 'day' );
-        $months = createSelectPrepare( range( 1, 12 ), 'Select Month' );
-        $self->createSelect( $months, 'month' );
-        $current_year = date( 'Y' );
-        $years = createSelectPrepare (
-            range( $current_year - $config[ 'age' ][ 'min' ], $current_year - $config[ 'age' ][ 'max' ] ),
-            'Select Year'
-        );
-        $self->createSelect( $years, 'year' );
```
dionyziz, line 56: Why are you removing this form?

controllers/user.php
```diff
@@ -74,7 +74,6 @@ public function update( $password = '', $password_new = '', $password_repeat = '
                 }
             }
             $user->email = $email;
-            $user->dateOfBirth = compact( 'day', 'month', 'year' );
```
dionyziz, line 77: Why are you removing the ability to set the dob?

PR #252, opened 2014-05-07
Throw error when specified method does not exist

### LINKED ISSUE #247: When invalid method is supplied, a 404 response should be returned
e.g. http://endofcodes.com/testrun/asdhufiopasidp

bugos:
Fixes #247.
I am not sure if the rewrite rule changes are should be included in this pull request.
Please review.

Before only requests in the form `endofcodes.com/recoursce/method` worked.
With the new RewriteRules the following requests are valid:

```
endofcodes.com/recource
endofcodes.com/recource/
endofcodes.com/recource/method
endofcodes.com/recource/method/
```

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -3,7 +3,6 @@ RewriteEngine on
 RewriteRule ^([a-z]+)/([a-z]+)/?$ index.php?resource=$1&method=$2 [L,QSA]
 RewriteRule ^([a-z]+)/?$ index.php?resource=$1 [L,QSA]
 
-
 <Files "run">
     Order Allow,Deny
     Deny from all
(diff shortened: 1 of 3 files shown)


.htaccess
```diff
@@ -1,6 +1,7 @@
 RewriteEngine on
 
-RewriteRule ^([a-z]+)/([a-z]+)$ index.php?resource=$1&method=$2 [L,QSA]
+RewriteRule ^([a-z]+)/([a-z]+)/?$ index.php?resource=$1&method=$2 [L,QSA]
```
dionyziz, line 3: Why do we have need the optional `/`?

PR #255, opened 2014-05-07
Define a new outputErrorPage() function in HTTPErrorExceptions

### LINKED ISSUE #254: HTTP exception must not echo HTML contents directly
Currently, HTTP exception classes output HTML content directly on construction. This does not separate concerns.

Suggest a method `$e->outputErrorPage()` on the exception class to output the error page, which will be called by the dispatcher.

bugos:
Fixes #254

### DIFF
--- a/run
+++ b/run
@@ -21,7 +21,7 @@
         exit( $controller->dispatch( [ 'method' => $method ], $vars, [], 'POST' ) );
     }
     catch ( HTTPErrorException $e ) {
-        die( 'An HTTP error occuρred: '. $e->header );
+        die( 'An HTTP error occurred: '. $e->header );
     }
     catch ( NotImplemented $e ) {
         die( 'An attempt was made to call a not implemented function: ' . $e->getFunctionName() . "\n" );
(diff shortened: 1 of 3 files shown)


run
```diff
@@ -20,6 +20,9 @@
     try {
         exit( $controller->dispatch( [ 'method' => $method ], $vars, [], 'POST' ) );
     }
+    catch ( HTTPErrorException $e ) {
+        die('An HTTP error occured: '. $e->header);
```
dionyziz: nit: spaces inside `( )`; you need a single space after `(` and a single space before `)`

dionyziz: :s/occured/occurred

vitsalis:
LGTM

run
```diff
@@ -20,6 +20,9 @@
     try {
         exit( $controller->dispatch( [ 'method' => $method ], $vars, [], 'POST' ) );
     }
+    catch ( HTTPErrorException $e ) {
+        die( 'An HTTP error occuρred: '. $e->header );
```
dionyziz: You seem to have an extraneous UTF-8 character `ρ` here.

run
```diff
@@ -20,6 +20,9 @@
     try {
         exit( $controller->dispatch( [ 'method' => $method ], $vars, [], 'POST' ) );
     }
+    catch ( HTTPErrorException $e ) {
+        die( 'An HTTP error occurred: '. $e->header );
```
dionyziz: nit: Space before `.`

PR #259, opened 2014-05-08
Fix line on player list

dionyziz:

PR #258, opened 2014-05-08
Add dropdown mockup

dionyziz:

PR #260, opened 2014-05-08
Add next,previous buttons and add css from mockup

dionyziz:


dionyziz:
Depends on #258 #259

views/game/view.php
```diff
@@ -105,8 +105,10 @@
     </div>
 </div>
 
-<div class='time'>
-    <span class='round'>Round <?php
+<div class='time' style="width:<?php
```
pkakelas: Maybe you should use jquery to set a css value in this div

dionyziz: Yes, perhaps. How do you suggest we provide the `$game->width` attribute to a jQuery context?

pkakelas: There are many ways to do that. My suggestion is to have a script tag inside view.php where the value will be assigned ta js variable like this "var name = <?php echo $game->width ?>".

dionyziz: I suggest we use a `data-width` attribute on a game-related HTML tag instead, or an `application/json` script tag which remains non-executable. It's good practice to separate data (as in this case) from executable code. This has the added security bonus of being forward-compatible with content-security-policy.

views/game/view.php
```diff
@@ -57,11 +57,11 @@
     ?></ol>
 </aside>
 
-<div class='game' style="width:<?php
+<div class='game' data-width="<?php
     echo $game->width * 20;
```
dionyziz: This should be the game's width, in number is available cells, not pixels. Similarly below.

vitsalis: ok then I close #271 , right?

dionyziz: No, you still need to have 20 somewhere, to convert from the number of cells to the number of pixels. This can be done at the javascript level, but you still need a constant.

dionyziz:
@pkakelas please review

views/game/view.php
```diff
@@ -105,8 +105,10 @@
     </div>
 </div>
 
-<div class='time'>
-    <span class='round'>Round <?php
+<div class='time' data-width="<?php
+    echo $game->width;
```
dionyziz: This has no place here. The data-width attribute is a semantic attribute indicating an internal property of the game. Don't use it as a presentation element indicating visual size. Use Javascript to read the game's semantic attributes and display every element you need appropriately.

static/script/game/view.js
```diff
@@ -3,6 +3,7 @@ var GameView = {
     ARROW_WIDTH: 30,
     roundCount: 0,
     maxHp: 0,
+    pixelMultiplier: 20,
```
dionyziz: nit: constant should be uppercase

static/script/game/view.js
```diff
@@ -105,7 +106,23 @@ var GameView = {
...
+        var $game = $( '.game' );
+        var width = $game.attr( 'data-width' );
+        var height = $game.attr( 'data-height' );
+
+        GameView.fixPlane( $game, {
+            "width": width,
+            "height": height
+        } );
+        GameView.fixPlane( $( '.time' ), {
+            "width": width
```
dionyziz: JSON in JS doesn't need quotes around keys

static/script/game/view.js
```diff
@@ -105,7 +106,23 @@ var GameView = {
...
+        } );
+    },
     ready: function() {
+        var $game = $( '.game' );
+        var width = $game.attr( 'data-width' );
+        var height = $game.attr( 'data-height' );
+
+        GameView.fixPlane( $game, {
+            "width": width,
+            "height": height
```
dionyziz: Similarly as below

static/script/game/view.js
vitsalis: yes but I have defined width

static/script/game/view.js
```diff
@@ -105,7 +106,23 @@ var GameView = {
         } );
         return false;
     },
+    fixPlane: function( $element, attributes ) {
+        $.each( attributes, function( index, value ) {
```
dionyziz: Rename `index` to `key`.

PR #262, opened 2014-05-08
Change output format if using json

dionyziz:

PR #264, opened 2014-05-08
CLI support for exporting all rounds of a game in JSON

dionyziz:


dionyziz:
Depends on #262

dionyziz:
@pkakelas please review

controllers/game.php
```diff
@@ -61,13 +61,20 @@ public function view( $gameid, $roundid = false ) {
...
             }
-            if ( isset( $this->acceptTypes[ 'application/json' ] ) ) {
-                require_once 'models/grader/serializer.php';
-                $mapJson = GraderSerializer::serializeCreatureList( $round->creatures );
-                require 'views/game/view.json.php';
-            }
-            else {
-                require 'views/game/view.php';
+            switch ( $this->outputFormat ) {
+                case 'json':
```
pkakelas, line 65: Are you sure that you want this case to be empty? Then why you set it?

vitsalis, line 65: it's not empty. I don't have a `break` here so if `outputFormat` is json, it executes the code from line 67.

PR #256, opened 2014-05-07
Implement the profile mockup

vitsalis:

PR #267, opened 2014-05-09
Make image upload work with new profile

vitsalis:


vitsalis:
Depends on #256

vitsalis:
When you hover over the image:

endofcodes_profile_5

dionyziz:
The "Upload an image" text should be vertically centered within the semi-transparent black overlay.

static/script/user/view.js
```diff
@@ -1,19 +1,87 @@
 var UserView = {
+    uploading: false,
+    UploadLinkOpacityMax: 0.8,
+    UploadLinkOpacityMin: 0.5,
```
dionyziz: Use UPPER_CASE_NAMES_SEPARATED_BY_UNDERSCORES for constants.

static/script/user/view.js
```diff
@@ -1,19 +1,87 @@
...
+        $( ".avatar img" ).remove();
+        $image = $( '<img src="' + source + '" alt="Profile Picture" />' );
+        $image.load( function() {
+            UserView.fixImageSize( $image );
+        } );
+        $( ".avatar" ).append( $image );
     },
     createImageError: function() {
-        $( '#image-form' ).prepend( "<div class='alert alert-danger'>This isn't an image</div>" )
+        $( '.text-center' ).prepend( "<div class='alert alert-danger'>This isn't an image</div>" )
```
dionyziz, line 14: Screenshot for this.

static/script/user/view.js
```diff
@@ -1,19 +1,87 @@
...
+            if ( makeOpacityBigger ) {
+                opacity = UserView.UploadLinkOpacityMax;
+            }
+            else {
+                opacity = UserView.UploadLinkOpacityMin;
+            }
+            UserView.fixUploadLinkOpacity( opacity );
+            setTimeout( function() {
+                UserView.animateImage( !makeOpacityBigger );
+            }, speed );
```
dionyziz: This animation is bad because:
- It discontinuously changes the opacity from min to max and back. The opacity should change smoothly, i.e. it should transition from min to max going through all the values in between.
- It is too fast
- It's not visible after you mouseout while the image is still uploading

I suggest you use CSS transitions to achieve a better effect - use one class that indicates the element should be animating and add it to the element using jQuery when the animation should be played. When the animation should be over, just remove the class. This will also allow you to remove the complexity from `finishUploadAnimation` and `startUploadAnimation` and just have them add/remove a CSS class. The `animateImage` function can then be fully moved to a couple of CSS rules.

static/script/user/view.js
dionyziz, line 14: "This isn't an image" always appears as an error. However, there are multiple erroneous cases: The image can be too big, the file can be a non-image, or the server may be misconfigured (i.e. there may be permission issues). We should distinguish between these.

static/script/user/view.js
```diff
@@ -38,20 +105,20 @@ var UserView = {
...
                 contentType: false,
-                success: function( res ) {
+                success: function( targetPath ) {
                     var reader = new FileReader();
 
+                    UserView.finishUploadAnimation();
+
                     reader.onloadend = function ( e ) {
-                        UserView.showUploadedImage( e.target.result );
+                        UserView.showUploadedImage( targetPath );
```
dionyziz, line 103: This is great.

static/script/user/view.js
dionyziz: Also suggest that you change the text to "Uploading..." while the upload is taking place.

static/script/user/view.js
```diff
@@ -1,19 +1,87 @@
...
+        }
     },
-    toggleSubmit: function() {
-        $( "#imageSubmit" ).toggle();
-        $( "#uploading" ).toggle();
+    fixUploadLinkOpacity: function( opacity ) {
+        $( "#upload-link" ).css( 'background-color', 'rgba(0,0,0,' + opacity + ')' );
+    },
+    animateImage: function( makeOpacityBigger ) {
+        var speed = 50;
```
dionyziz: nit: `SPEED`

static/script/user/view.js
```diff
@@ -1,19 +1,87 @@
...
+        $( ".avatar" ).mouseover( function() {
+            if ( $( ".profile-header" ).attr( 'data-sameUser' ) == 'yes' ) {
+                $( "#upload-link" ).show();
+            }
+        } );
+        $( ".avatar" ).mouseout( function() {
+            $( "#upload-link" ).hide();
+        } );
+        $( "#upload-link" ).click( function() {
+            $( "#image" ).trigger( 'click' );
```
dionyziz, line 62: You should not allow the user to upload an image while an image is already uploading. If the user uploads a bunch of images in a row, the last one is not currently kept, and it therefore constitutes a bug. In addition, if a user is allowed to upload another image while the first image is uploading (and the avatar in the page has not been replaced), they feel that their previous action was not correctly registered. You should check the `uploading` variable at this point.

static/script/user/view.js
```diff
@@ -1,19 +1,87 @@
...
     removeImageError: function() {
-        $( '#image-form .alert.alert-danger' ).remove();
+        $( '.text-center .alert.alert-danger' ).remove();
+    },
+    fixImageSize: function( $image ) {
+        var height, width;
+        var imgWidth = $image.width();
+        var imgHeight = $image.height();
+
+        height = width = 168;
```
dionyziz: Define as CONSTANTS.

views/user/view.php
```diff
@@ -2,74 +2,77 @@
...
+            echo "no";
+        }
+        else {
+            echo "yes";
+        }
+    ?>">
+        <div class="avatar">
+            <a href="" id="upload-link">Upload an image</a>
+            <img src="<?php
+                echo $user->image->target_path;
```
dionyziz: You should display something more user-friendly than a failing image when the user hasn't uploaded an avatar.

static/style/user/view.css
```diff
@@ -1,3 +1,111 @@
-#uploading {
+.avatar {
+    width: 168px;
+    height: 168px;
```
dionyziz, line 3: Maybe it's time to switch to SASS for CSS to ensure DRY. @pkakelas @VitSalis your opinion on this?

static/script/user/view.js
vitsalis, line 14: I think we should do this in another branch

dionyziz, line 14: OK

static/style/user/view.css
vitsalis, line 3: It seems more useful, so yes.

views/user/view.php
vitsalis: Like what?

dionyziz: A default image with a shadow of a human face like Facebook does?

static/script/user/view.js
vitsalis, line 62: so, should I hide the upload-link and in it's place put some text like "Uploading.." ?

dionyziz, line 62: Or change its text content

vitsalis, line 62: yes but if you click that, the image selector will be triggered again.

vitsalis:
error_image

vitsalis:
Default picture: 

pkakelas_profile

static/script/user/view.js
dionyziz, line 62: So have an `uploading` boolean check there.

static/script/user/view.js
```diff
@@ -1,19 +1,92 @@
...
+            }
+            else {
+                opacity = UserView.UPLOAD_LINK_OPACITY_MIN;
+            }
+            UserView.fixUploadLinkOpacity( opacity );
+            setTimeout( function() {
+                UserView.animateImage( !makeOpacityBigger );
+            }, UserView.SPEED );
+        }
+    },
```
dionyziz: Use CSS transitions.

static/script/user/view.js
```diff
@@ -1,36 +1,70 @@
 var UserView = {
+    uploading: false,
+    UPLOAD_LINK_OPACITY_MAX: 0.8,
+    UPLOAD_LINK_OPACITY_MIN: 0.5,
```
dionyziz: Unnecessary

static/script/user/view.js
```diff
@@ -1,36 +1,70 @@
...
+    IMAGE_HEIGHT: 168,
+    IMAGE_WIDTH: 168,
+    SPEED: 50,
```
dionyziz: Unnecessary

static/script/user/view.js
```diff
@@ -1,36 +1,70 @@
...
-            $image.width( width );
+            $image.width( UserView.IMAGE_WIDTH );
             $image.css( 'left', 0 );
-            $image.css( 'top', -Math.floor( ( $image.height() - height ) / 2 ) );
+            $image.css( 'top', -Math.floor( ( $image.height() - UserView.IMAGE_HEIGHT ) / 2 ) );
         }
+    },
+    finishUploadAnimation: function() {
+        $( ".uploading" ).hide();
+        $( ".uploading" ).removeClass( 'animate' );
```
dionyziz, line 36: upload-link is not reshown, user is unable to reupload avatar

static/script/user/view.js
```diff
@@ -1,36 +1,70 @@
...
+        $( ".uploading" ).show();
+        $( ".uploading" ).addClass( 'animate' );
+        UserView.uploading = true;
+    },
+    ready: function() {
+        $( '.avatar img' ).load( function() {
+            UserView.fixImageSize( $( '.avatar img' ) );
+        } );
+        $( ".avatar" ).mouseover( function() {
+            if ( $( ".profile-header" ).attr( 'data-sameUser' ) != 'yes' || UserView.uploading ) {
```
dionyziz: Prefer `== 'no'` instead of `!= 'yes'`

static/style/user/view.css
```diff
@@ -91,6 +88,38 @@ ul.contact li {
...
+    padding-top: 5px;
+    position: absolute;
+    text-shadow: 1px 1px black;
+    color: white;
+    background-color: rgba( 0, 0, 0, 0.5 );
+}
+.uploading.animate {
+    -webkit-animation: pulse 1s infinite;
+}
+@-webkit-keyframes pulse {
```
dionyziz, line 113: Browser compatibility

static/style/user/view.css
```diff
@@ -91,6 +88,38 @@ ul.contact li {
...
+    0% {
+        background-color: rgba(0,0,0,0.5);
+    }
+    50% {
+        background-color: rgba(0,0,0,0.8);
+    }
+    100% {
+        background-color: rgba(0,0,0,0.5);
```
dionyziz, line 121: Your style is not consistent with above

static/script/user/view.js
vitsalis, line 36: `UserView.uploading = false` on the next line means that on the next `mouseover` the upload-link will be reshown.

dionyziz, line 36: Yes, you're right.

PR #268, opened 2014-05-09
rename google+ to google

### LINKED ISSUE #211: Login page: Google+ login should be Google login
We should aim to support login with Google, not Google+. Let's change the text in the login form.

pkakelas:
fixes #211


vitsalis:
LGTM

PR #272, opened 2014-05-10
Home mockup

pkakelas:
screenshot from 2014-05-10 18 02 12


dionyziz:
Left-align Google button with username input

dionyziz:
Right-align GitHub button with username input

dionyziz:
Left- and right- align submit button with username input

dionyziz:
Center Google and github text within buttons - exclude icon part from the width within which you align

mockups/home.html
```diff
@@ -0,0 +1,205 @@
...
+                </li>
+                <li>
+                <a class="btn btn-block btn-social btn-github">
+                    <i class="fa fa-github"></i>Github
+                </a>
+                </li>
+                </ul>
+            </div>
+        </div>
+        <div id='last'>
```
vitsalis: last? I don't think that's a good id.

mockups/home.html
```diff
@@ -0,0 +1,205 @@
...
+                        <td>
+                            <a href="user/view?username=vitsalis">vitsalis</a>
+                        </td>
+                        <td>Greece</td>
+                        <td>Coming soon</td>
+                    </tr> 
+                </tbody>
+            </table> 
+        </div>
+        <div id='screenshot'>
```
vitsalis: I don't think `screenshot` is a good id. You should better give it an id like `prevGame`

pkakelas:
I find submit button more beautiful as it is. If you think its better the other way I can change it.

pkakelas:
The leftmost pixel of google button has the same x with the username input. The rightmost pixel of github button has the same x with the username input. I don't understand what else I should do.

dionyziz:
No, they don't.

ee5414d2-d859-11e3-98ee-0bbb81e40eb0

dionyziz:
Why do you need the file mockups/google.png?

pkakelas:
screenshot from 2014-05-13 17 32 32

dionyziz:
Still need to center text within submit buttons.

dionyziz:
Postponing this until we pair-program redesign this together w/ @pkakelas.

PR #294, opened 2014-05-15
Make form errors appear before a specific input in forms and add default focus

pkakelas:
depends on #291


static/script/error.js
```diff
@@ -1,4 +1,6 @@
-function createError( form, description ) {
+function createFormError( form, inputName, description, focus ) {
+    focus = focus || true;
     $( '.alert' ).remove();
-    $( form ).prepend( "<div class='alert alert-danger'>" + description + "</div>" );
+    $( "<div class='alert alert-danger'>" + description + "</div>" ).insertBefore( form + ' input[name=' + inputName + ']' );
+    $( '[name=' + inputName + ']', form ).focus();
```
dionyziz: you should also `.select()`

dionyziz:
What happens if I submit the form once with one error and then again with a different error? Shouldn't the original error be hidden?

pkakelas:
@dionyziz this is why I use $( '.alert' ).remove();

PR #298, opened 2014-05-16
Remove repo icons from footer

dionyziz:


pkakelas:
Don't you think we must replace it with something else?

dionyziz:
I will replace it in a subsequent branch with developer information (profiling) for now. For users, we should move the blog and rules links to the bottom.

dionyziz:
screen shot 2014-05-17 at 5 51 25 pm

PR #299, opened 2014-05-16
Introduce user roles

dionyziz:


dionyziz:
This is needed to add profiling abilities to developers. It will also help guard permissions for executing migrations and running unit tests.

database/migration/20140516235600_add_user_default_role.php
```diff
@@ -0,0 +1,4 @@
+<?php
```
pkakelas, line 1: Niice, I've never thought that we can use the migrate function independently

models/user.php
```diff
@@ -152,6 +155,9 @@ protected function onBeforeSave() {
             if ( !isset( $this->boturl ) ) {
                 $this->boturl = '';
             }
+            if ( $this->role < 0 || !isset( $this->role ) ) {
```
pkakelas, line 158: How can this happen?

dionyziz, line 158: Negative values cannot happen unless they are set specifically; this is just a safeguard. Unset roles can occur if a new user is constructed and no role is assigned.

pkakelas, line 158: why don't you make $this-role a boolean?

dionyziz, line 158: For future extensibility; we may want to add other roles inbetween user and developer, e.g. "admin" or "contributor" that don't have developer roles, but have some increased user abilities.

PR #302, opened 2014-05-17
Refactor footer view.

dionyziz:


dionyziz:
Depends #299, #298

static/style/footer.css
```diff
@@ -29,3 +29,6 @@ body {
 #main {
     margin-bottom: 65px;
 }
+#footer .dev {
```
pkakelas: Where is the class "dev" being assigned ?

PR #305, opened 2014-05-17
Optimize game performance

vitsalis:


models/game.php
```diff
@@ -45,8 +46,13 @@ public function __construct( $id = false ) {
                 $countrounds = $data[ 'countrounds' ];
                 if ( $countrounds > 0 ) {
                     $this->rounds[ 0 ] = new Round( $this, 0 );
-                    for ( $i = 1; $i < $countrounds; ++$i ) {
-                        $this->rounds[ $i ] = new Round( $this, $i, $this->rounds[ 0 ] );
+                    if ( $roundid && $roundid != 0 ) {
```
dionyziz: The left-hand-side of the `&&` also checks for non-zero ints, because PHP converts to bool to check an `if` condition. Did you mean to check `$roundid !== false` instead of just `$roundid`?

models/round.php
```diff
@@ -37,38 +38,59 @@ public function __construct( $a = false, $b = false, $genesis = false ) {
...
+
+                $creaturesInfo = dbSelect(
                     'roundcreatures',
                     [ 'creatureid', 'action', 'direction', 'hp', 'locationx', 'locationy' ],
                     compact( 'roundid', 'gameid' )
                 );
+
                 // map: creatureid => user
+                $creatureToUser = [];
+                // map: user => user
```
dionyziz: `userid => user`

models/round.php
```diff
@@ -37,38 +38,59 @@ public function __construct( $a = false, $b = false, $genesis = false ) {
...
-                        $user_info = dbSelectOne(
-                            'creatures',
-                            [ 'userid' ],
-                            compact( 'id', 'gameid' )
-                        );
-                        $user = new User( $user_info[ 'userid' ] );
+                        $id = $creatureInfo[ 'creatureid' ];
+
+                        foreach ( $usersInfo as $userInfo ) {
+                            if ( $userInfo[ 'id' ] == $id ) {
```
dionyziz: Preprocess the `$userInfo` array to turn it into a dict. It makes for more readable code (and as a bonus, it is faster too)

models/game.php
```diff
@@ -27,7 +28,7 @@ public static function getLastGame() {
             return new Game( $game[ 0 ][ 'id' ] );
         }
 
-        public function __construct( $id = false ) {
+        public function __construct( $id = false, $roundid = false ) {
```
dionyziz, line 31: You must test the alternative way of calling the `Game` constructor.

dionyziz:
Please provide performance metrics before/after the optimization and experimental setting.

vitsalis:
Performance metrics about game with Width(14) Height(15):
- Without optimization : Time 0.19562888145447
- With optimization      : Time 0.015771865844727

The performance is optimized when there is a `roundid` attribute on `$_GET`. It is expected that the user will only experince the not optimized time only once worst case. While browsing the game with the arrows the performance is optimized.

Issue #314, opened 2014-05-20
Avatar image temporarily appears wrongly sized during image upload


dionyziz:
When uploading an avatar image, after the upload finishes, the image appears temporarily in the wrong size while it loads until it is properly resized and cropped by the Javascript. Make it only appear in the correct size.

PR #319, opened 2014-05-21
Fix image size before loading it

vitsalis:


vitsalis:
Fixes #314

dionyziz:
merge master

static/script/user/view.js
```diff
@@ -23,11 +25,13 @@ var UserView = {
 
         if ( imgWidth > imgHeight ) {
             $image.height( UserView.IMAGE_HEIGHT );
+            $image.width( Math.floor( UserView.IMAGE_HEIGHT * imgWidth / imgHeight ) );
```
dionyziz, line 28: This seems to work, but why is it needed? The browser should calculate the appropriate width, right?

static/script/user/view.js
```diff
@@ -3,12 +3,14 @@ var UserView = {
     IMAGE_HEIGHT: 168,
     IMAGE_WIDTH: 168,
     showUploadedImage: function( source ) {
-        $( ".avatar img" ).remove();
         $image = $( '<img src="' + source + '" alt="Profile Picture" />' );
         $image.load( function() {
+            $image.width( this.width );
+            $image.height( this.height );
```
dionyziz, line 9: I'm confused as to why you're setting these. Are you setting the width to what it already is?

PR #320, opened 2014-05-21
Add name, surname, website and github info

pkakelas:


dionyziz:
fix build

dionyziz:
There is a lot of duplication in this pull request. We should discuss how to eliminate some of it. There are abstractions to be made in the controller, the dispatcher, the model, the error handling. An association between controller and model may be appropriate in cases like this. Model attributes may need to be dynamically generated from the attribute list and then only white-listed on the controller.

tests/models/user.php
```diff
@@ -9,11 +9,19 @@ public function testCreate() {
...
+            $user->name = 'Dimitris';
+            $user->surname = 'Lamprinos';
+            $user->github = 'https://github.com/pkakelas';
+            $user->website = 'https://pkakelas.com';
             $user->save();
             $passwordSuccess = $user->authenticatesWithPassword( 'secret1234' );
             $this->assertTrue( $passwordSuccess, 'Password must be the one associated during creation' );
             $this->assertEquals( 'pkakelas', $user->username, 'Username must be the one associated during creation' );
             $this->assertEquals( 'pkakelas@gmail.com', $user->email, 'Email must be the one associated during creation' );
+            $this->assertEquals( 'Dimitris', $user->name, 'Name must be the one associated during creation' );
```
bugos, line 21: What's the point of checking `$user->name` that you have set exactly above. Are we verifying that it's written in database correctly with this assertion?

pkakelas, line 21: I make this check in order to assert that after the save function the attributes of the user object are not changing.

tests/models/user.php
```diff
@@ -9,11 +9,19 @@ public function testCreate() {
...
+            $this->assertEquals( 'Lamprinos', $user->surname, 'Surname must be the one associated during creation' );
+            $this->assertEquals( 'https://github.com/pkakelas', $user->github, 'Github must be the one associated during creation' );
+            $this->assertEquals( 'https://pkakelas.com', $user->website, 'Website must be the one associated during creation' );
         }
```
vitsalis, line 25: You should also check if `$_SESSION[ 'user' ]` is set and is the same as `$user`

views/user/create.php
```diff
@@ -78,7 +78,24 @@
...
+            $self->createError( 'Surnames can contain only letters' );
+        }
+        $self->createLabel( 'surname', 'Surname' );
+        $self->createInput( 'text', 'surname' );
+        if ( isset( $website_invalid ) ) {
+            $self->createError( 'Please enter a valid website' );
+        }
+        $self->createLabel( 'website', 'Website' );
+        $self->createInput( 'text', 'website' );
+        $self->createLabel( 'github', 'Github username' );
```
vitsalis: no checks for github? What are we going to do with a github username? We have to check if it's a valid one.

tests/models/user.php
dionyziz, line 25: I don't think it's appropriate to test for `$_SESSION` in a model test; models should be session-agnostic.

vitsalis, line 25: you're right.

dionyziz:
merge master

tests/models/user.php
```diff
@@ -43,6 +51,30 @@ public function testEmailChange() {
...
+            $this->assertEquals( 'https://pkakelas.com', $user->website, 'Website must be the one associated during update' );
+        }
+        public function testGithubChange() {
+            $user = $this->buildUser( 'pkakelas' );
+            $user->github = 'https://github.com/pkakelas';
+            $user->save();
+            $this->assertEquals( 'https://github.com/pkakelas', $user->github, 'Github must be the one associated during update' );
+        }
         public function testSetCountry() {
             $country = $this->buildCountry( 'Greece', 'GR' );
```
dionyziz, line 80: Test website validaton?

views/user/create.php
```diff
@@ -78,7 +78,24 @@
...
+        $self->createInput( 'text', 'github' );
```
dionyziz: I think these shouldn't be part of the registration form. Let's keep registration short. User update can include them.

models/user.php
```diff
@@ -130,10 +140,12 @@ protected function onBeforeSave() {
                 }
                 throw new ModelValidationException( 'password_small' );
             }
+            if ( !empty( $this->website ) && !filter_var( $this->website, FILTER_VALIDATE_URL ) ) {
+                    throw new ModelValidationException( 'website_invalid' );
```
dionyziz: nit: indent

tests/models/user.php
```diff
@@ -50,6 +82,35 @@ public function testSetCountry() {
...
+                'The website URL string must have "http://" in the beginning'
+            );
+            $this->assertThrows(
+                function() {
+                    $user = $this->buildUser( 'pkakelas' );
+                    $user->website = "http://pkakelas123.com";
+                    $user->save();
+                },
+                'ModelValidationException',
+                'The website URL cannot contain numbers'
```
dionyziz: I think just one assertion is enough - we are checking that the user model uses URL validation. The URL validation feature of PHP itself is already tested, so we don't need to be exhaustive here.

views/user/create.php
```diff
@@ -23,9 +23,9 @@
 <?php
     $form = new Form( 'user', 'create' );
     $form->id = 'register-form';
-    $form->output( function( $self ) use( $username_empty, $username_invalid, $password_empty,
-            $email_empty, $username_used, $password_small,
-            $password_not_matched, $email_used, $email_invalid, $countries, $location ) {
+    $form->output( function( $self ) use( $username_empty, $username_invalid, $password_empty, 
+            $email_empty, $username_used, $password_small, $password_not_matched, $email_used, $email_invalid,
+            $countries, $location ) {
```
dionyziz: align below variables above

controllers/user.php
```diff
@@ -59,7 +64,8 @@ public function view( $username, $image_invalid ) {
         }
 
         public function update( $password = '', $password_new = '', $password_repeat = '',
-                                $countryShortname = '', $email = '', $day = '', $month = '', $year = '' ) {
+                                $countryShortname = '', $email = '', $day = '', $month = '', $year = '', $name,
+                                $surname, $website, $github ) {
```
dionyziz, line 68: Optional arguments must optionally be followed by optional arguments only.

controllers/user.php
```diff
@@ -1,7 +1,8 @@
 <?php
     class UserController extends AuthenticatedController {
         public function create( $username = '', $password = '', $password_repeat = '', $email = '',
-                                $countryShortname = '', $day = '', $month = '', $year = '' ) {
+                                $countryShortname = '', $day = '', $month = '', $year = '', $name = '', $surname = '',
+                                $website = '', $github = '' ) {
```
dionyziz, line 5: Let's remove `$website`, `$github`, `$name`, `$surname` from create.

vitsalis:
This PR is stalling

PR #331, opened 2014-05-27
Test that user is logged in after registration.

dionyziz:


models/lib/simple_html_dom.php
```diff
@@ -0,0 +1,1721 @@
+<?php
```
dionyziz, line 1: This is the third party library simplehtmldom. No need to review this.

helpers/http.php
```diff
@@ -35,7 +38,9 @@ public function __construct( $resource_or_url = false, $method = false, $args =
                     }
                     $args[ $key ] = "$key=" . urlencode( $arg );
                 }
-                $this->url = "$resource/$method?" . implode( "&", $args );
+                $this->method = $method;
+                $this->args = $args;
+                $this->url = $this->resource . '/' . $this->method . '?' . implode( "&", $args );
```
vitsalis: `$this->args` ?

dionyziz: This looks fine to me

helpers/http.php
```diff
@@ -25,7 +28,7 @@ public function __construct( $resource_or_url = false, $method = false, $args =
                 $this->url = $resource_or_url;
             }
             else {
-                $resource = $resource_or_url;
+                $this->resource = $resource_or_url;
```
bugos: How about we rename `$resource_or_url` to `$resourceOrUrl` throughout the file since we are using CamelCase?

models/lib/simple_html_dom.php
```diff
@@ -0,0 +1,1721 @@
...
+    function __toString()
+    {
+        return $this->root->innertext();
+    }
+
+    function __get($name)
+    {
+        switch ($name)
+        {
+            c
```
bugos, line 1695: New line at the end of file.

helpers/http.php
bugos: We should rename to `$arguments`.

Also `'&'` instead of `"&"`.

controllers/testrun.php
```diff
@@ -1,11 +1,14 @@
 <?php
     class TestrunController extends ControllerBase {
-        protected $environment = 'test';
+        public $environment = 'test';
 
         public function create( $name, $all = false ) {
-            set_time_limit( 360 );
+            if ( $all ) {
```
bugos: The `$all` argument is not needed. We set the default value for `$name = ''` and then call `create()` without arguments to run all tests.

models/test/functional.php
```diff
@@ -0,0 +1,74 @@
...
+
+    class FunctionalTestRequest {
+        protected $unittest;
+        public $resource;
+        public $method;
+        public $verb;
+        public $vars;
+        public $session;
+
+        public function __construct( UnitTest $unittest, $resource, $method, $session, $verb, $vars ) {
```
bugos: `UnitTest $unitTest`

models/test/functional.php
```diff
@@ -0,0 +1,74 @@
...
+        public $content;
+        public $dom;
+
+        public function __construct( UnitTest $unittest, $content ) {
+            $this->content = $content;
+            $this->unittest = $unittest;
+            $this->dom = str_get_html( $content );
+        }
+        public function assertHas( $selector, $description = '' ) {
+            $this->unittest->assertTrue( ( bool )$this->dom->find( $selector ), $description );
```
bugos: Better use something more descriptive like `$CssSelector`.

I think `( bool )` is not needed.

helpers/http.php
dionyziz: Why rename it to `$arguments`? I think `$args` is a clear abbreviation. Generally, this idea is also followed by the PHP language: Write out a word fully, unless its shortening is standard and clear.

In programming, shortening "argument" to "arg" is quite traditional: See [argc and argv](http://crasseux.com/books/ctutorial/argc-and-argv.html) in the C language and [$argc](http://www.php.net/manual/en/reserved.variables.argc.php) and [$argv](http://www.php.net/manual/en/reserved.variables.argv.php) in PHP. Even in build-in PHP functions, the shortening of "argument" to "arg" is typically used, and also the shortening of "arguments" to "args". For example, see [func_get_arg](http://docs.php.net/manual/en/function.func-get-arg.php), [func_get_args](http://docs.php.net/manual/en/function.func-get-args.php), [func_num_args](http://docs.php.net/manual/en/function.func-num-args.php). The use of "args" as a shortening for "arguments" is typical even within [standard examples](http://www.php.net/manual/en/function.func-num-args.php#example-5208) in PHP documentation.

While it is not necessary for us to endorse these practices, I don't see why we shouldn't. I don't believe making your recommended change here increases code maintainability or clarity.

models/lib/simple_html_dom.php
dionyziz, line 1695: I don't think we should modify this file. As it's an external library, leaving it completely unmodified will make it easier to upgrade it later in our project lifecycle if there are updates pushed to the original repository. If we make small changes like these, we should maintain patches that should be applied to future updates, which is cumbersome. Since they are not completely necessary, I recommend that we maintain this file as-is and we don't make any stylistic changes. I suggest we maintain the missing new line from end-of-file from the original author.

models/test/functional.php
dionyziz: The word "selector" is standard in computer science nomenclature, especially within the web development field. See [what Wikipedia has to say about selector](https://en.wikipedia.org/wiki/Selector) to verify that the only legitimate use within web development cycles is the one intended and also [google selector](https://www.google.com/search?q=selector) to see that the first three results indeed confirm this usage. I prefer to avoid `$CssSelector`, as this is a generic XML selector and in reality constitutes an XPath selector, but using the name `XPathSelector` makes this somewhat obscure. I recommend that we maintain the name `$selector`.

controllers/testrun.php
```diff
@@ -1,11 +1,16 @@
...
     class TestrunController extends ControllerBase {
-        protected $environment = 'test';
+        public $environment = 'test';
 
-        public function create( $name, $all = false ) {
-            set_time_limit( 360 );
+        public function create( $name = '' ) {
+            $all = $name == '';
+
+            if ( $all ) {
```
bugos: I suggest that we merge this `if` statement with the one on line 16 and check directly if `$name == ''`. I don't think that the `$all` variable makes the code more readable.

models/lib/simple_html_dom.php
bugos, line 1695: Agreed.

bugos:
LGTM

helpers/http.php
```diff
@@ -35,7 +38,9 @@ public function __construct( $resource_or_url = false, $method = false, $args =
                     }
                     $args[ $key ] = "$key=" . urlencode( $arg );
                 }
-                $this->url = "$resource/$method?" . implode( "&", $args );
+                $this->method = $method;
+                $this->args = $args;
+                $this->url = $this->resource . '/' . $this->method . '?' . implode( '&', $args );
```
vitsalis: You use `this` for `resource` and for `method`. why not for `args`?

models/test/functional.php
```diff
@@ -0,0 +1,74 @@
...
+            $this->method = $method;
+            $this->verb = $verb;
+            $this->vars = $vars;
+            $this->session = $session;
+        }
+        public function execute() {
+            $oldSession = $_SESSION;
+            try {
+                ob_start();
+                $controller = controllerBase::findController( $this->resource );
```
vitsalis: Capital `c` on `controllerBase`

tests/controllers/user.php
```diff
@@ -0,0 +1,19 @@
...
+                'username'         => 'dionyziz',
+                'password'         => 'secret1234567',
+                'password_repeat'  => 'secret1234567',
+                'email'            => 'dionyziz@gmail.com',
+                'countryShortname' => 'GR',
+                'day'              => 30,
+                'month'            => 11,
+                'year'             => 1987
+            ] );
+            $response->assertHas( 'a[href="user/view?username=dionyziz"]', 'User must be logged in after registration' );
```
vitsalis: this tests if there is a link to dionyziz's profile, not that the user is logged in.

tests/controllers/user.php
```diff
@@ -0,0 +1,19 @@
+<?php
+    class UserControllerTest extends FunctionalTest {
+        public function testCreate() {
+            $response = $this->request( 'user', 'create', 'POST', [
+                'username'         => 'dionyziz',
+                'password'         => 'secret1234567',
+                'password_repeat'  => 'secret1234567',
+                'email'            => 'dionyziz@gmail.com',
+                'countryShortname' => 'GR',
```
vitsalis, line 9: You only need `username`, `password`, `password_repeat`, and `email` for a valid request.

dionyziz, line 9: Why does it hurt to have a registration test with full details? I'll add more tests to make sure certain fields can be omitted later.

tests/controllers/user.php
dionyziz: That's the way the user _perceives_ that they are logged in - they look at the navbar and see a link to their profile. As a functional test, this treats the website as a black box, submits requests in the user's stead, and examines the results from the user's point of view.

tests/controllers/user.php
vitsalis, line 9: it doesn't, it was just a suggestion

tests/controllers/user.php
vitsalis: well you're looking at the whole website, not the navbar.

vitsalis:
LGTM should I merge?

PR #334, opened 2014-05-29
Additional tests for user controller

dionyziz:


dionyziz:
Depends #331

tests/controllers/user.php
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        /* create */
+        private function makeCreateRequest() {
+            $this->response = $this->request( 'user', 'create', 'POST', $this->userDetails );
+        }
+        private function assertValidates( $fields, $error, $description ) {
+            $this->userDetails = array_merge( $this->userDetails, $fields );
+            $this->makeCreateRequest();
+            $this->response->assertContains( $error, $description . ". Expected error \"$error\" to be produced, but validation incorrectly passed" );
```
bugos, line 26: Why don't you use dots in the end of your assertion descriptions?

dionyziz, line 26: This is our style throughout the whole test suite across the codebase. If you like, you can modify `UnitTestFailedException` to add a dot to every description.

dionyziz:
@pkakelas @VitSalis please review

dionyziz:
Pair programming w/ @VitSalis

models/test/functional.php
```diff
@@ -59,16 +63,44 @@ public function execute() {
...
+                throw new DOMInvalidException( $this );
+            }
+            $this->unitTest->assertTrue( strpos( $this->dom->save(), $text ) !== false, $description );
+        }
+        public function assertStatusIs( $status, $description = '' ) {
+            $this->unitTest->assertEquals( $status, $this->status, $description . ". Expected $status, found " . $this->status . "." );
+        }
+    }
+
+    class DOMInvalidException extends Exception {
```
pkakelas, line 94: What is this Exception for?

tests/models/geolocation.php
```diff
@@ -1,5 +1,24 @@
 <?php
     require_once 'models/geolocation.php';
+    require_once 'models/network.php';
```
pkakelas, line 3: This changes should not be in this pull request. Pls change the name of the pull request or create a new one.

models/test/functional.php
dionyziz, line 94: This exception is thrown by the functional test framework in case the DOM of the page that was returned by the controller is invalid XML and is not parseable, therefore the DOM cannot be constructed. In that case, it is impossible to run selector-based assertions such as `assertHas`.

tests/models/geolocation.php
dionyziz, line 3: This is not unrelated. The controller test for the user calls the registration page; as the call is internal, geolocation may fail unexpectedly. In particular, if running unit tests offline, we wish to at least have the user controller tests succeed. It's necessary to therefore introduce indication of failure in geolocation, which in turn needs to be tested. Testing it requires these mocks and changes. How do you propose we split it up?

PR #339, opened 2014-06-07
Fix image upload bug

vitsalis:


vitsalis:
When you're not logged in the "upload an image" message appears in every profile.
endofcodes

PR #341, opened 2014-06-08
Use absolute paths on gamescript

dionyziz:


dionyziz:
Pair programming w/ @VitSalis

Issue #257, opened 2014-05-08
HTTPError pages not loaded correctly due to wrong base.


bugos:
Depends on #252.
The config is not loaded on HTTPError pages (like 404), so the base is set to "".
The problem occures on pages like `endofcodes.com/invalid/invalid`.

bugos:
I propose that we make the `ControllerBase->getConfig()` function static and we call it from index.php.

vitsalis:
I think we should make an http errors controller. Every controller passes from ControllerBase so we wouldn't have this problem. Also it will be a better API.

bugos:
@VitSalis that's an interesting idea but it would require a lot of refactoring. And I am not sure of how we would call the error controller after we have caught an exception in `index.php`.

I am looking at the code in `ControllerBase` and I think that the function `loadConfig()` does not suit there. Only the last line does load the config. The rest of the code determines the enviroment.

I propose that we make a new function `findEnv()` in `/helpers/config.php/`  and move the enviroment code there. Then we can easily load the config file from `index.php` as I proposed above.

vitsalis:
Everything passes from controllers except from http errors at this point. I think that's wrong.
The ControllerBase gives the basis for a lot of things that are used throughout the whole page.

Also there won't be a lot of refactoring. Just an `http.php` controller, with a `create` method, that takes the HTTP error as a parameter and includes the corresponding view.

bugos:
Ok I agree; I will fix this with a new controller.

@VitSalis thanks for the feedback.

PR #351, opened 2014-06-13
Error controller to handle HTTP errors.

### LINKED ISSUE #310: Redirects with go() don't work in CLI.
The `go()` function is defined in `helpers/http.php` and is used to cleanly switch between controllers with an HTTP redirect. 

However this method does not work in CLI.

Additionally it has the disadvantages that: 
1. the original URL is ~~lost~~ changed and 
2. any error parameters have to be passed as GET parameters.

bugos:
Fixes #310 and #257.

In this pull request:
- I defined a new `ErrorRedirectException` that  can be used instead of `RedirectException` to change the controller withoud reloading the page with an HTTP redirect. 
- I changed the `HTTPErrorException` to extend the new `ErrorRedirextException`. When thrown it changes the controller to the new `HTTPErrorController` as @VitSalis suggested in #257.
- I refactored the `ControllerBase`.

### DIFF
--- a/controllers/base.php
+++ b/controllers/base.php
@@ -5,6 +5,7 @@
         public $trusted = false;
         public $outputFormat = 'html';
         public $pageGenerationBegin; // time marking the beginning of page generation, in epoch seconds
+        protected $method = 'view'; // Override to specify a default controller method.
 
         public static function findController( $resource ) {
             $resource = basename( $resource );
@@ -17,45 +18,6 @@
             $controller = new $controllername();
 
             return $controller;
-        }
-        protected function protectFromForgery( $token = '', $httpRequestMethod = '' ) {
-            if ( $httpRequestMethod === 'POST'
-            && ( !isset( $_SESSION[ 'form' ] )
-              || !isset( $_SESSION[ 'form' ][ 'token' ] )
-              || $token !== $_SESSION[ 'form' ][ 'token' ]
-              || $token == '' )
-            && !$this->trusted ) {
-                throw new HTTPUnauthorizedException( 'Your CSRF token was invalid.' );
-            }
-        }
-        protected function getControllerMethod( $requestedMethod, $httpRequestMethod ) {
-            $method = $requestedMethod;
-
-            try {
-                if ( Form::getRESTMethodIdempotence( $method ) === 1 && $httpRequestMethod != 'POST' ) {
-                    $method .= 'View';
-                }
-            }
-            catch ( HTMLFormInvalidException $e ) {
-                $method = 'view';
-            }
-
-            return $method;
-        }
-        protected function getControllerVars( $get, $post, $files, $httpRequestMethod ) {
-            switch ( $httpRequestMethod ) {
-                case 'POST':
-                    $vars = array_merge( $post, $files );
-                    break;
-                case 'GET':
-                    $vars = $get;
-                    break;
-                default:
-                    $vars = [];
-                    break;
-            }
-
-            return $vars;
         }
         protected function sessionCheck() {
             global $config;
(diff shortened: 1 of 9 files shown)


controllers/base.php
```diff
@@ -57,25 +52,43 @@ protected function getControllerVars( $get, $post, $files, $httpRequestMethod )
...
+                $this->method = $vars[ 'method' ];
+            }
+            try {
+                $idempotence = Form::getRESTMethodIdempotence( $this->method ) === 1 ;
+            }
+            catch ( HTMLFormInvalidException $e ) {
+                throw new HTTPNotFoundException( $this->method . ' is not a valid REST method.' );
+            }
+            if ( $idempotence && $httpRequestMethod != 'POST' ) {
+                    $this->method .= 'View';
```
dionyziz: nit: indent

controllers/base.php
```diff
@@ -57,25 +52,43 @@ protected function getControllerVars( $get, $post, $files, $httpRequestMethod )
...
-                $_SESSION[ 'user' ] = $user;
             }
         }
-        protected function callWithNamedArgs( $methodReflection, $callable, $vars ) {
+        private function getControllerMethod( $vars, $httpRequestMethod ) {
+            if ( isset( $vars[ 'method' ] ) ) {
+                $this->method = $vars[ 'method' ];
+            }
+            try {
+                $idempotence = Form::getRESTMethodIdempotence( $this->method ) === 1 ;
```
dionyziz: nit: remove space before `;`

dionyziz: Don't `===` compare against `1`. `1` is truthy and can be used directly in an `if` condition (as is appropriate in a weakly typed language like PHP). But if you prefer, modify `getRESTMethodIdempotence` to return `true` or `false` instead.

controllers/base.php
```diff
@@ -1,10 +1,11 @@
 <?php
     abstract class ControllerBase {
-        protected $acceptTypes = [];
+        public $acceptTypes = [];
```
dionyziz: Why `public`? Would the dispatcher ever have need to modify or read these?

helpers/http.php
```diff
@@ -45,7 +45,7 @@ public function __construct( $resourceOrURL = false, $method = false, $args = []
         }
     }
 
-    class HTTPErrorException extends Exception {
+    class HTTPErrorException extends ErrorRedirectException {
```
dionyziz, line 76: I'm not sure how semantically an `HTTPErrorException` is-an `ErrorRedirectException`. For example, 404 is not a redirect.

index.php
```diff
@@ -14,15 +14,7 @@
...
-
-        $url = $e->getURL();
-
-        header( 'Location: ' . $config[ 'base' ] . $url );
-    }
-    catch ( HTTPErrorException $e ) {
-        header( $e->header );
-        $e->outputErrorPage();
+    catch ( ErrorRedirectException $e ) {
+        $e->callErrorController(); //should catch exception inside the ControllerBase object.
```
dionyziz: You must also modify the other dispatcher - run.

controllers/httperror.php
```diff
@@ -0,0 +1,11 @@
+<?php
+    class HTTPErrorController extends ControllerBase {
+      public $method = 'create';
```
dionyziz: nit: indentation

controllers/httperror.php
```diff
@@ -0,0 +1,11 @@
+<?php
+    class HTTPErrorController extends ControllerBase {
+      public $method = 'create';
+
+        public function createView( $error, $reason, $header ) {
```
dionyziz: Why are we using the `create` RESTful semantics? I know @VitSalis suggested this, but it doesn't seem that the user is asking for an error to be created here - only viewing an error that has occurred.

helpers/http.php
bugos, line 76: Would it be better if I throwed an ErrorRedirectException inside the HTTPErrorException constructor?

dionyziz, line 76: Why do you need to do that? It's not a redirect.

bugos, line 76: The point is that when an `HTTPErrorException` (including `404`) is thrown we want to change(redrect) the controller to the new `HTTPErrorController`. This is done when an ErrorRedirectExceprtion is catched with a `$e->controller` set to `httperror`.

run
```diff
@@ -8,23 +8,26 @@
...
+        }
+        catch ( HTTPErrorException $e ) {
+            die( 'An HTTP error occurred: ' . $e->header . ': ' . $e->reason );
+        }
+        catch ( ErrorRedirectException $e ) {
+            launchController( $e->controller, $e->arguments );
+        }
     }
+
+    exit( launchController( $resource, [ 'method' => $method ], $vars, [], 'POST' ) );
```
dionyziz, line 32: Why do you need an `exit` here?

bugos, line 32: So that the return value of the `run` script is the return value of the dispatcher. This is how it was before; should we  change it?

dionyziz, line 32: No, this is fine, I was confused.

helpers/html.php
```diff
@@ -131,10 +131,10 @@ public static function getRESTMethodIdempotence( $method ) {
                 'update' => 1,
                 'view' => 0
             ];
-            if ( isset( $methods[ $method ] ) ) {
-                return $methods[ $method ];
+            if ( !isset( $methods[ $method ] ) ) {
+                throw new HTMLFormInvalidException( $method );
             }
-            throw new HTMLFormInvalidException( $method );
+            return $methods[ $method ];
```
dionyziz, line 137: Nice touch.

helpers/http.php
```diff
@@ -1,4 +1,24 @@
...
+                list( $a, $q ) = explode( ';q=', $a );
+                if ( $q === 0 ) {
+                    continue;
+                }
+            }
+            $acceptTypes[ $a ] = true;
+        }
+        return $acceptTypes;
+    }
+
```
dionyziz, line 21: Good refactoring to move this to the separate HTTP helper :+1:

dionyziz, line 21: We should unit test this. Do you want to do this in this pull request or merge and work on it separately?

PR #353, opened 2014-06-14
Add an abstract environment 'defaults' in config for DRY

### LINKED ISSUE #249: config file does not follow DRY between environments
The config file currently contains a lot of duplication between the `test` and `development` environments. Suggest that we devise a mechanism to make an environment copy its configuration from another environment and then override particular settings.

bugos:
Fixes #249 and #289.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -27,14 +27,12 @@
             ],
             'forgot_password_exp_time' => 3600 * 24,
             'cli_max_width' => 120,
-            'email' => 'team@endofcodes.com',
-            'root' => '/var/www/',
-            'base' => 'http://localhost/endofcodes/'
+            'email' => 'team@endofcodes.com'
         ],
         'development' => [
             'db' => [
                 'user' => 'endofcodes',
-                'pass' => 'sample_pass',
+                'pass' => 'sample_pass', //override in config-local.php
                 'dbname' => 'endofcodes'
             ],
 
--- a/config/config.php
+++ b/config/config.php
@@ -1,6 +1,6 @@
 <?php
     return [
-        'development' => [
+        'defaults' => [
             'db' => [
                 'host' => 'localhost',
                 'user' => 'endofcodes',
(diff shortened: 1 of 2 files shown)


config/config.php
```diff
@@ -27,36 +27,24 @@
             ],
             'forgot_password_exp_time' => 3600 * 24,
             'cli_max_width' => 120,
-            'email' => 'team@endofcodes.com'
+            'email' => 'team@endofcodes.com',
+            'root' => '/var/www/',
+            'base' => 'http://localhost/endofcodes/'
```
dionyziz: Why do `root` and `base` need to be specified since they can be deduced?

bugos: I think we should keep all used config keys in config.php. This way config.php can be used as a reference for our config system.

helpers/config.php
```diff
@@ -9,21 +9,29 @@ function getBase() {
...
     }
+    function calculateConfigEntries( &$config ) {
+        $config[ 'root' ] = getcwd();
+        if ( isset( $_SERVER[ 'HTTP_HOST' ] ) ) {
+            $config[ 'base' ] = getBase();
+        }
+    }
+    function validateConfigEntries( $config ) {
+        if ( $config[ 'development' ][ 'db' ][ 'dbname' ] == $config[ 'test' ][ 'db' ][ 'dbname' ] ) {
+            throw new ModelValidationException( " Database name for development and testing is the same." );
```
dionyziz, line 22: nit: Remove space from the beginning of quotes

config/config.php
dionyziz: Yes, but we want config-local to not contain a `root` and `base` key and for them to be derived automatically. But this way they will default to these values and they will need to be overwritten. I think these keys in particular should never be required to be specified in any config files – they are helpful to have for our application to function, but are not really part of the configuration formally.

helpers/config.php
bugos, line 22: It is required for the error to be displayed properly.

bugos, line 22: btw config should be model

PR #357, opened 2014-08-07
[GET_ON_BOT_INITIALIZE] Replace the POST request on /bot with a GET request

vitsalis:


dionyziz:
Please change the spec.

vitsalis:
https://github.com/dionyziz/endofcodes/blob/master/SPECIFICATION.md#initiation-1

I don't think we need to change it. It already says "GET request"

models/grader/bot.php
```diff
@@ -78,7 +78,7 @@ protected function httpRequest( $endpoint = '', $method = 'view', $data = array(
         }
         public function sendInitiateRequest() {
             try {
-                $ch = $this->httpRequest( 'bot', 'create' );
+                $ch = $this->httpRequest( 'bot', 'view' );
```
dionyziz, line 83: Please test that this is actually a `GET` request.

PR #358, opened 2014-08-09
Node bot

vitsalis:


bots/node/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "multer": "^0.1.3",
+    "mocha": "^1.4.14",
+    "request": "^2.39.0"
+  },
+  "devDependencies": {},
+  "scripts": {
+    "test": "mocha"
+  },
+  "author": "Vitalis Salis",
+  "license": "ISC"
```
dionyziz: Please release your code under the MIT license.

dionyziz:
Your tests should be automated and included in travis.

bots/node/lib/bot.js
```diff
@@ -0,0 +1,113 @@
...
+} );
+
+app.post( '/round', function( req, res ) {
+    function randomDirection( array ) {
+        var directions = [ 'NORTH', 'EAST', 'SOUTH', 'WEST' ];
+        return directions[ array[ Math.floor( Math.random() * array.length ) ] ];
+    }
+
+    var round = req.body.round;
+    var map = JSON.parse( req.body.map );
```
petrosagg, line 54: This can throw if `req.body.map` isn't well formed and will take down the server with it. Needs a `try{} catch {}` and an appropriate response code to the user (probably 400)

bots/node/lib/bot.js
```diff
@@ -0,0 +1,113 @@
...
+
+app.listen( 8000 );
+
+app.get( '/bot', function( req, res ) {
+    res.send( JSON.stringify( {
+        botname: 'sample_botname',
+        version: '0.1.0',
+        username: 'sample_username'
+    } ) );
+    res.status( 200 ).end();
```
petrosagg: You don't need to specify `status` or call `end()` when you've used the `send()` method

bots/node/lib/bot.js
```diff
@@ -0,0 +1,113 @@
...
+} );
+
+app.post( '/game', function( req, res ) {
+    res.send( JSON.stringify( {} ) );
+    res.status( 200 ).end();
```
petrosagg: Same here

bots/node/lib/bot.js
```diff
@@ -0,0 +1,113 @@
...
+            var attacks = [], moves = [];
+            var offsets = [
+                [ 0, 1 ],
+                [ 1, 0 ],
+                [ 0, -1 ],
+                [ -1, 0 ]
+            ];
+
+            for ( var j = 0; j < offsets.length; ++j ) {
+                var newX = parseInt( x ) + offsets[ j ][ 0 ], newY = parseInt( y ) + offsets[ j ][ 1 ];
```
petrosagg: It's a good practice to pass a second argument to `parseInt` which is the base of the number you're trying to parse. This is to prevent inputs like '0312' to be treated as octal numbers. i.e `parseInt(x, 10)` instead of `parseInt(x)`

bots/node/lib/bot.js
```diff
@@ -0,0 +1,113 @@
...
+                    } );
+                }
+            }
+        }
+    }
+
+    res.send( JSON.stringify( {
+        'intent': intent
+    } ) );
+    res.status( 200 ).end();
```
petrosagg: Same here for `status` and `end`

bots/node/lib/bot.js
```diff
@@ -0,0 +1,113 @@
...
+    extended: true
+} ) );
+
+app.use( multer( {
+    dest: './uploads/'
+} ) );
+
+app.listen( 8000 );
+
+app.get( '/bot', function( req, res ) {
```
petrosagg: I'd suggest that API calls be prefixed with an API version like `/v1/bot` so that it's clear when you change it and/or make backwards incompatible changes.

bots/node/lib/bot.js
```diff
@@ -0,0 +1,113 @@
...
+    } ) );
+    res.status( 200 ).end();
+} );
+
+app.post( '/game', function( req, res ) {
+    res.send( JSON.stringify( {} ) );
+    res.status( 200 ).end();
+} );
+
+app.post( '/round', function( req, res ) {
```
petrosagg: This function uses user input without any sort of validation. Any trivially malformed input will cause a `TypeError` exception when trying to read properties from undefined values which will again take the whole server down with it.

bots/node/lib/bot.js
```diff
@@ -0,0 +1,113 @@
...
+
+            if ( attacks.length ) {
+                intent.push( {
+                    'creatureid': creature.creatureid,
+                    'action': 'ATTACK',
+                    'direction': randomDirection( attacks )
+                } );
+            }
+            else {
+                if ( !moves.length || Math.random() * 2 < 1 ) {
```
petrosagg: Why not `Math.random() < 0.5` ?

bots/node/lib/bot.js
```diff
@@ -0,0 +1,113 @@
...
+} ) );
+
+app.use( multer( {
+    dest: './uploads/'
+} ) );
+
+app.listen( 8000 );
+
+app.get( '/bot', function( req, res ) {
+    res.send( JSON.stringify( {
```
petrosagg: You can use the `res.json()` method and `express` will handle the serialisation for you.

petrosagg: Same for all other methods in this file

bots/node/lib/bot.js
petrosagg, line 54: An other note here is that, because you're using the `bodyParser` middleware, if the POST request is sent with the correct Content-Type header (i.e `application/json`) then `req.body.map` will already be an object because `bodyParser` parses it for you. I think the best approach to this is testing whether the `Content-Type` header is `application/json` and if not return a 400 status code without further processing.

bots/node/lib/bot.js
petrosagg: Plus, it will correctly set the response header to `application/json`

bots/node/test/bot.js
```diff
@@ -0,0 +1,311 @@
...
+            .end( done );
+    } );
+} );
+
+describe( 'round request', function() {
+    function requestAndCheckResponse( map, validDirections, validActions, myid, numCreatures, done ) {
+        //console.log( JSON.parse( JSON.stringify( map ) ) );
+        agent
+            .post( '/round' )
+            .send( 'map=' + JSON.stringify( map ) +'&myid=1&W=10&H=10&round=1&gameid=1' )
```
petrosagg: This `JSON.stringify( map )` should probably be wrapped in `encodeURIComponent()` to avoid weird errors.

bots/node/test/bot.js
```diff
@@ -0,0 +1,311 @@
...
+            .expect( function( res ) {
```
petrosagg, line 51: Add these two before the expect call?

```
  .set('Accept', 'application/json')
  .expect('Content-Type', /json/)
```

bots/node/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "main": "lib/bot.js",
+  "directories": {
+    "test": "test"
+  },
+  "dependencies": {
+    "body-parser": "^1.6.0",
+    "express": "^4.8.0",
+    "multer": "^0.1.3",
+    "mocha": "^1.4.14",
+    "request": "^2.39.0"
```
dionyziz, line 13: Please include `supertest`.

dionyziz:
Node best practices recommend a README file. Please include a simple one. A one-line description is OK.

dionyziz:
mocha tests are passing for this PR, ready to merge test-wise.

bots/node/lib/bot.js
```diff
@@ -0,0 +1,116 @@
...
+        var directions = [ 'NORTH', 'EAST', 'SOUTH', 'WEST' ];
+        return directions[ array[ Math.floor( Math.random() * array.length ) ] ];
+    }
+
+    var round = req.body.round;
+    try {
+        var map = JSON.parse( req.body.map );
+    }
+    catch ( e ) {
+        res.status( 400 ).end();
```
petrosagg: You can use `res.send(400)`

pkakelas:
lgtm

bots/node/lib/bot.js
dionyziz: For reference, `res.send(400)` is deprecated, so the patch looks good. Will review further tonight.

bots/node/lib/bot.js
dionyziz, line 54: Requires #359; will merge for now, we will work on this on a separate PR.

dionyziz:
Test automation postponed for a separate PR as of #361; will merge this for now.

bots/node/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "description": "A sample bot in node for EndOfCodes",
+  "main": "lib/bot.js",
+  "directories": {
+    "test": "test"
+  },
+  "dependencies": {
+    "body-parser": "^1.6.0",
+    "express": "^4.8.0",
+    "multer": "^0.1.3",
+    "mocha": "^1.4.14",
```
dionyziz, line 16: Test requirements should be in `devDependencies`.

bots/node/test/bot.js
```diff
@@ -0,0 +1,349 @@
...
+}
+
+describe( 'bot request', function() {
+    it( 'should return the bot credentials', function( done ) {
+        agent
+            .get( '/bot' )
+            .expect( function( res ) {
+                var expData = '{"botname":"sample_botname","version":"0.1.0","username":"sample_username"}';
+
+                assert.strictEqual( expData, res.text );
```
dionyziz: Your assertions should not check for exact serialized equality, but that the JSON objects are deeply equal. This is the bot contract.

bots/node/test/bot.js
```diff
@@ -0,0 +1,349 @@
...
+describe( 'game request', function() {
+    it( 'should return an empty object', function( done ) {
+        agent
+            .post( '/game' )
+            .set( 'Accept', 'application/json' )
+            .expect( 'Content-Type', /json/ )
+            .expect( function( res ) {
+                var expData = '{}';
+
+                assert.strictEqual( expData, res.text );
```
dionyziz: Similar.

bots/node/test/bot.js
```diff
@@ -0,0 +1,349 @@
...
+            } )
+            .end( done );
+    } );
+} );
+
+describe( 'round request', function() {
+    function requestAndCheckResponse( map, validDirections, validActions, myid, numCreatures, done ) {
+        agent
+            .post( '/round' )
+            .send( 'map=' + encodeURIComponent( JSON.stringify( map ) ) +'&myid=1&W=10&H=10&round=1&gameid=1' )
```
dionyziz: Use dict.

bots/node/test/bot.js
```diff
@@ -0,0 +1,349 @@
...
+            .set( 'Accept', 'application/json' )
+            .expect( 'Content-Type', /json/ )
```
dionyziz: DRY

bots/node/test/bot.js
```diff
@@ -0,0 +1,349 @@
...
+                assert.strictEqual( intent[ 1 ].action, 'ATTACK' );
+                assert.strictEqual( intent[ 1 ].direction, 'SOUTH' );
+            } )
+            .end( done );
+    } );
+
+    it( 'should respond with a 400 status code when invalid json is sent', function( done ) {
+        agent
+            .post( '/round' )
+            .send( 'map={fadfas&myid=1&W=10&round=1&gameid=1' )
```
dionyziz: Dict.

bots/node/test/bot.js
```diff
@@ -0,0 +1,349 @@
...
+                x: 1,
+                y: 2,
+                hp: 100
+            },
+            {
+                creatureid: 3,
+                userid: 3,
+                x: 2,
+                y: 1,
+                hp: 100
```
dionyziz, line 158: Have some sane defaults, e.g. `hp` is often `100` unless you overwrite it. DRY.

bots/node/lib/bot.js
```diff
@@ -0,0 +1,121 @@
...
+
+app.listen( 8000 );
+
+function tryFn( fn ) {
+    return function( req, res ) {
+        try {
+            fn( req, res );
+        }
+        catch ( e ) {
+            res.status( 400 ).end();
```
petrosagg: This should be 500

bots/node/test/bot.js
```diff
@@ -0,0 +1,372 @@
...
+var assert = require( 'assert' );
+var app = require( '../lib/bot.js' );
+var agent = request.agent( app );
+
+describe( 'bot request', function() {
+    it( 'should return the bot credentials', function( done ) {
+        agent
+            .get( '/v1/bot' )
+            .expect( function( res ) {
+                var expData = {"botname":"sample_botname","version":"0.1.0","username":"sample_username"};
```
dionyziz: nit: style

PR #374, opened 2014-08-16
Automation of ids in every body

pkakelas:


views/header.php
```diff
@@ -47,7 +47,10 @@
         ?>
         <meta charset="utf-8" />
     </head>
-    <body>
+    <?php
+        $id = "body-{$this->resource}-" . str_replace( 'View', '', $this->method  );
+    ?>
+    <body id=<?= $id ?>>
```
dionyziz: XSS

dionyziz: Need quotes around id

controllers/base.php
```diff
@@ -18,6 +18,7 @@ public static function findController( $resource ) {
             require_once $filename;
             $controllername = ucfirst( $resource ) . 'Controller';
             $controller = new $controllername();
+            $controller->resource = $resource;
```
dionyziz, line 22: `resource` must be defined in controller as a public attribute

PR #379, opened 2014-09-10
Disable dbconfig in production

bugos:
The dbconfig controller should not be accessible in production as it exposes the database credentials and allows any user to change them.

### DIFF
--- a/controllers/dbconfig.php
+++ b/controllers/dbconfig.php
@@ -4,6 +4,10 @@
         protected function dbInit() {}
 
         public function create( $user, $pass, $dbname ) {
+            if ( $this->environment == 'production' ) {
+                go(); // disable in production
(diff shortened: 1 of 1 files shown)


bugos:
Please review

controllers/dbconfig.php
```diff
@@ -4,6 +4,10 @@ class DbconfigController extends ControllerBase {
         protected function dbInit() {}
 
         public function create( $user, $pass, $dbname ) {
+            if ( $this->environment == 'production' ) {
+                go(); // disable in production
+            }
+
```
dionyziz, line 10: This can be abstracted into `ControllerBase` as `$this->disableInProduction()` or `$this->developmentOnly()`.

PR #382, opened 2015-05-17
Changed DBConfigController to run all available migrations.

nikatlas:


controllers/dbconfig.php
```diff
@@ -1,13 +1,14 @@
 <?php
     class DbconfigController extends ControllerBase {
         protected $method = 'create';
-        protected function dbInit() {}
+        // protected function dbInit() {} // deprecated dunno why this was here.
```
dionyziz, line 4: This was here because the `dbInit()` function is called by all controllers through ControllerBase, so it must be substituted for a dummy one if the database isn't set up. @bugos can tell you more.

controllers/dbconfig.php
```diff
@@ -1,13 +1,14 @@
...
         protected $method = 'create';
-        protected function dbInit() {}
+        // protected function dbInit() {} // deprecated dunno why this was here. 
 
         public function create( $user, $pass, $dbname ) {
             $entries = compact( 'user', 'pass', 'dbname' );
             $entries = [ 'db' => $entries ];
             try {
                 updateConfig( $entries, $this->environment );
+                $this->runMigrations();
```
dionyziz, line 11: I don't think this script was ever meant to run migrations, only to write a config file. Perhaps it should take you to a different controller to do that, or have a more generic installation script? I agree that it would be good to do that, but this is not the appropriate place to put it.

controllers/dbconfig.php
```diff
@@ -28,6 +29,20 @@ public function createView( $error, $dbSaid ) {
...
+            foreach ( $migrations as $migration ){
+                try {
+                    require_once 'database/migration/' . $migration;
+                }
+                catch( MigrationException $ex ){
+                    // Maybe change this behaviour to revert changes till now or something...
+                    throw $ex;
+                }
+            }
+        }
```
dionyziz, line 45: This function should not be part of the dbConfig controller, which is responsible for writing the database configuration file. Perhaps the migration controller is more appropriate?

nikatlas:
Would it be ok to redirect to /migrationrun/create 
POSTING all=1?
Haven't checked the code just a thought

dionyziz:
Sounds like a possibly good idea, I'd like to hear the opinions of @bugos
and @pkakelas who have worked on this.

bugos:
I think we should just display a warning (`There are pending migrations`) and a link to migrationrun/create where the user can run the migrations by himself, so that he knows what is going on.

controllers/dbconfig.php
bugos, line 4: We can add a comment here: `Prevent db initialisation while config is updated`

nikatlas:
Seems legit.

dionyziz:
OK let's go with that.

controllers/dbconfig.php
dionyziz, line 4: @bugos Please make a PR.

